PR #1, opened 2023-08-17
make work

Peyton-McKee:
work mac


RChandler234 (changes requested):
Could you actually remove those two binary files and add them to the gitignore, I don't think it's great practice to store binary files in github. I'll test that it still works on mine without them (they should hopefully j be generated with npx prisma generate

Issue #4, opened 2023-09-10
Migrate From NodeJS to BunJS


Peyton-McKee:
Bun is more relevant for the application we are using as it has a web socket framework built into it and allows for 7x more throughput than NodeJS.

Peyton-McKee:
see Linked PR #5

PR #5, opened 2023-09-11
#4 Bunn

Peyton-McKee:
Bun bun bun bun

Backend is in bun but it keeps crashing
I dont understand why
I dont think its prisma because wouldnt the backend just not load then
just try docker-compose build 
and then docker-compose up and see what happens


Peyton-McKee:
bun has too little documentation and not enough docker support for us to use unfortunately maybe in the future

Issue #15, opened 2023-09-11
CRUD Operations - Get All Systems


Peyton-McKee:
### Description

We need an endpoint to get all the systems in the database

### Acceptance Criteria

- [ ] Service function which returns all the systems from the database with their runs
- [ ] Unit Testing (maybe)

### Proposed Solution

^
epic #8

PR #21, opened 2023-09-13
#15 - CRUD get all systems

leoleader:
## Changes

added service function for get all systems

## Notes

currently just returning result from findMany as json, not entirely sure if that's the desired
way to return resulting data for this so would like feedback

## Test Cases

- Haven't decided on unit testing for this yet

## To Do

- [] Unit testing (maybe)


scylla-server/src/index.ts
```diff
@@ -20,6 +20,18 @@ app.get("/runs", async (req: Request, res: Response) => {
   }
 });
 
+// #15 CRUD Operations 
+// Service Function, gets all systems and returns result as json
+app.get("/systems", async (req: Request, res: Response) => {
```
Peyton-McKee, line 25: ok so this would be great if we were still using http. But now that were using a socket we actually just need the prisma.system.findMany() call in an exported function in  a system.services.ts file in a services folder.

Peyton-McKee (changes requested):

scylla-server/src/index.ts
leoleader: ok i've updated so it should work for websocket

scylla-server/src/services/systems.services.ts
```diff
@@ -0,0 +1,15 @@
...
+// CRUD operation to get all systems with ResponseFunction type
+
+export const getAllSystems: ResponseFunction = async (data: JSON) => {
+  try {
+    const data = await prisma.system.findMany();
+    // was getting type errors trying to return just data
+    // so added JSON parse and stringify to make it conform to JSON
+    return JSON.parse(JSON.stringify(data));
+  } catch (error) {
+    throw new Error('Error getting all systems');
```
Peyton-McKee, line 13: you shouldnt need a try catch here

scylla-server/src/utils/message-maps.utils.ts
```diff
@@ -7,7 +11,7 @@ export const createServerMessageMap = (): Map<string, ResponseFunction> => {
   const serverMessageMap = new Map<string, ResponseFunction>();
   serverMessageMap.set('test', (data: JSON) => {
     console.log('test', data);
-    return data;
+    return Promise.resolve(data);
```
Peyton-McKee, line 14: why did you add this?

scylla-server/src/utils/message-maps.utils.ts
```diff
@@ -20,7 +24,10 @@ export const createClientMessageMap = (): Map<string, ResponseFunction> => {
   const clientMessageMap = new Map<string, ResponseFunction>();
   clientMessageMap.set('test', (data: JSON) => {
```
Peyton-McKee, line 25: you can delete this now, it was just a template for what it would look like

scylla-server/src/services/systems.services.ts
```diff
@@ -0,0 +1,15 @@
+import prisma from '../prisma/prisma-client';
+import { ResponseFunction } from '../utils/message-maps.utils';
+
+// CRUD operation to get all systems with ResponseFunction type
+
```
Peyton-McKee, line 5: use JavaDoc i.e /** */ it should prompt you with paramter values when you do the autocomplete with VSCode

scylla-server/src/utils/message-maps.utils.ts
```diff
@@ -20,7 +24,10 @@ export const createClientMessageMap = (): Map<string, ResponseFunction> => {
...
     console.log('test', data);
-    return data;
+    return Promise.resolve(data);
   });
+  // adds the getAllSystems CRUD to ClientMessageMap
+  clientMessageMap.set('getAllSystems', getAllSystems);
+  console.log('getAllSystems', getAllSystems);
```
Peyton-McKee, line 31: you can delete this console log

scylla-server/src/services/systems.services.ts
```diff
@@ -0,0 +1,15 @@
...
+export const getAllSystems: ResponseFunction = async (data: JSON) => {
```
Peyton-McKee, line 6: since were not using data in this particular function (its undefined) you can preface it like this _data so that the linting thing goes away

scylla-server/src/services/systems.services.ts
```diff
@@ -0,0 +1,15 @@
...
+  try {
+    const data = await prisma.system.findMany();
+    // was getting type errors trying to return just data
+    // so added JSON parse and stringify to make it conform to JSON
+    return JSON.parse(JSON.stringify(data));
```
Peyton-McKee, line 11: ok i didnt realize you would have to do this, so instead can you change the ResponseFunction type to return a string rather than a JSON Object and then just return JSON.stringify(data)

Peyton-McKee (changes requested):
Lit Job! Left a few comments.

scylla-server/tests/services.test.ts
```diff
@@ -0,0 +1,16 @@
+import prisma from '../src/prisma/prisma-client';
```
Peyton-McKee, line 1: can you specify this file even more to be systems-services.test.ts

Peyton-McKee (approved):
WWWWW

Issue #32, opened 2023-09-19
Scylla Server - Receive Siren Message


Peyton-McKee:
### Description

Every message we receive from Siren is going to be in the form of ServerMessage from OdysseyBase, create a function that parses this message in that format and then loop through all the data and use the upsertData function to propogate it to the database

### Acceptance Criteria

- [x] Parses data as servermessage from the proxy client
- [x] #55
- [ ] if its the first time the data has been received since connecting to the car, upsert a new run
- [ ] upsert all the nodes as you receive them
- [ ] loop through all the data and upsert it to the database with the given time

### Proposed Solution

- [ ] Code in the proxy-client class

PR #57, opened 2023-10-02
#32 receive-scylla-server-message

leoleader:
## Changes

- handleMessage method in proxyClient
- handleData method in proxyClient
- createNewRun: boolean and currentRun: Run properties added to proxyClient
- gonna need to update odyssey lol
- added even more stuff!

## Notes

- tbh i havent looked at this ticket in a WHILE and argos has changed a good bit since i made the first draft so while i think this makes sense it could be completely off so review and comments are very much appreciated
- i do not know how to test this lol

Closes # 32


scylla-server/src/proxy/proxy-client.ts
```diff
@@ -39,26 +61,108 @@ export default class ProxyClient {
...
+      const clientdata: ClientData = {
+        name: serverdata.name,
+        value: serverdata.value,
+        units: serverdata.units,
+        timestamp: data.unix_time
+      };
+      clientData.push(clientdata);
+
+      switch (serverdata.name) {
+        case 'driverUser':
```
RChandler234, line 129: should be enum values, not raw strings

scylla-server/src/proxy/proxy-client.ts
```diff
@@ -39,26 +61,108 @@ export default class ProxyClient {
...
+          driverName = serverdata.value as unknown as string;
+          break;
+        case 'systemName':
+          systemName = serverdata.value as unknown as string;
```
RChandler234, line 133: can you not just do as string?

scylla-server/src/proxy/proxy-client.ts
```diff
@@ -39,26 +61,108 @@ export default class ProxyClient {
...
    * @param message The message received from Siren
    */
-  private handleMessage = (topic: string, payload: Buffer) => {};
-
+  private handleMessage = (topic: string, payload: Buffer) => {
+    // am i supposed to do something with topic
+    console.log('Received Message: ', payload);
+    try {
+      const data = JSON.parse(payload.toString()) as ServerMessage;
+      this.handleData(data);
```
RChandler234, line 80: so my one issue with this right now is that if one part of handleData fails, the whole thing fails, and all the code is built upon that assumption, I think ideally we should future proof this a bit, because eventually we will wanna be able to salvage parts of a message even if other parts of it are wrong

scylla-server/src/proxy/proxy-client.ts
```diff
@@ -39,26 +61,108 @@ export default class ProxyClient {
...
+    }
+
+    if (systemName) {
+      SystemService.upsertSystem(systemName, this.currentRun.id);
+    }
+    if (driverName) {
+      DriverService.upsertDriver(driverName, this.currentRun.id);
+    }
+    if (latitude && longitude && radius && locationName) {
+      LocationService.upsertLocation(locationName, latitude, longitude, radius, this.currentRun.id);
```
RChandler234, line 160: should also be continuously tracking latitude/ longitude as a dataType - Peyton

scylla-server/src/proxy/proxy-client.ts
```diff
@@ -39,26 +61,108 @@ export default class ProxyClient {
...
+    let radius = undefined;
+    let locationName = undefined;
+    let driverName = undefined;
+    let systemName = undefined;
+
+    // iterating and upserting
+    const clientData: ClientData[] = [];
+    for (const serverdata of data.data) {
+      // transform serverdata into client data to send to ProxyServer
+      const clientdata: ClientData = {
```
RChandler234, line 120: like right here, we just send all the data to the frontend regardless of if the data type, system, location, etc has been correctly added to the database. I would move this adding to the clientData to the bottom of the for loop

RChandler234 (changes requested):

scylla-server/src/proxy/proxy-client.ts
leoleader: in the build it yells at me because its possible type conversion from num to string, and says to add as unknown first

scylla-server/src/proxy/proxy-client.ts
leoleader: can u elaborate on this a lil, i made it so that latitude and longitude are upserted as dataTypes

scylla-server/src/proxy/proxy-client.ts
leoleader: for being able to salvage parts of the message, what if we changed handleData so that it was just the stuff inside the for loop and then pulled out the rest into handleMessage, then handleMessage would iterate through the ServerData and call handleData in a try catch for each ServerData. That way if handleData fails for one ServerData, the whole thing doesn't fail.

angular-client/src/services/socket.service.ts
```diff
@@ -24,6 +24,7 @@ export default class SocketService {
     this.socket.on('message', (message: string) => {
       try {
         const data = JSON.parse(message) as ServerData;
+        storage.setCurrentRunId(data.runId);
```
RChandler234, line 28: so storage is just acting as a cache?

angular-client/src/services/socket.service.ts
```diff
@@ -24,6 +24,7 @@ export default class SocketService {
     this.socket.on('message', (message: string) => {
       try {
```
RChandler234, line 25: there should be some more comments throughout this method so it's easier to follow what lines are doing what

scylla-server/src/index.ts
```diff
@@ -56,9 +56,12 @@ const serverSocket = new Server(server, {
   }
 });
 
+const proxyServers: ProxyServer[] = [];
```
RChandler234: can I get some context on what these proxyServers are for?

scylla-server/src/proxy/proxy-client.ts
```diff
@@ -1,21 +1,48 @@
...
+import SystemService from '../odyssey-base/src/services/systems.services';
+import ProxyServer from './proxy-server';
+import { ClientData } from '../utils/message.utils';
+import { ServerData, ServerMessage } from '../odyssey-base/src/types/message.types';
 /**
  * Handler for receiving messages from Siren
  */
 export default class ProxyClient {
   connection: MqttClient;
+  // should a new run be created, based off if new connection & data received
```
RChandler234: did you mean to phrase this as a question?

scylla-server/src/proxy/proxy-client.ts
```diff
@@ -1,21 +1,48 @@
...
+  newLocation: boolean = true;
 
   /**
    * Constructor
    * @param socket The socket to send and receive messages from
    */
-  constructor(mqttClient: MqttClient) {
+  constructor(mqttClient: MqttClient, proxyServers: ProxyServer[]) {
     this.connection = mqttClient;
+    // only true first time after connected and data received
```
RChandler234: these comments aren't great. They should be clearer/ be full thoughts if they're gonna be here

scylla-server/src/proxy/proxy-client.ts
```diff
@@ -39,29 +67,148 @@ export default class ProxyClient {
...
    * @param message The message received from Siren
    */
-  private handleMessage = (topic: string, payload: Buffer) => {
-    //TODO: Handle the message
-    console.log('Received Message: ', topic, payload.toString());
-  };
+  private handleMessage = (topic: string, payload: Buffer, packet: IPublishPacket) => {
+    try {
+      const data = JSON.parse(payload.toString()) as ServerData;
+      /* Infer node name from topics first segment */
```
RChandler234, line 80: like these, love these comments

scylla-server/src/utils/message.utils.ts
```diff
@@ -8,9 +8,10 @@ export type ClientMessage = {
 /**
  * The format of the data sent to the client
  */
-type ClientData = {
+export type ClientData = {
```
RChandler234, line 11: this type is here, and in socket-utils in the frontend as 'ServerData.' If there's an easy way to avoid duplication that'd be nice, but not a huge deal

RChandler234 (changes requested):

angular-client/src/services/socket.service.ts
Peyton-McKee: Its also where we access all the most recent real time data points

scylla-server/src/index.ts
Peyton-McKee: Theyre all the connections to the server. So if u connect and then i connect both connections will have their owjnprody server

scylla-server/src/proxy/proxy-client.ts
Peyton-McKee: That was one of dylans comments i forgot to delete

scylla-server/src/utils/message.utils.ts
Peyton-McKee: Its like if we had a shared library the packages are independent of each other so we either make a shared folder or just make the types twice

angular-client/src/services/socket.service.ts
RChandler234: receiveData

RChandler234 (approved):

Issue #52, opened 2023-10-02
CRUD Operation - Get All Runs routes and controllers


Peyton-McKee:
### Description

We need an endpoint to view all the possible runs

### Acceptance Criteria

- [ ] Route and controller for get all runs

### Proposed Solution

look at finishline
epic #8

Peyton-McKee:
closed via #47

Peyton-McKee:
wrong ticket

Issue #50, opened 2023-10-02
CRUD - Route and Controller for datatypes


Peyton-McKee:
### Description

We need to allow http request for getting all datatypes from the database and also getting data by datatype name

### Acceptance Criteria

- [ ] Route and controller for get all datatypes
- [ ] Route and controller for get data by datatype name

### Proposed Solution

Look at finishline
epic #8

PR #62, opened 2023-10-10
#49 routes and controllers for CRUD

### LINKED ISSUE #49: CRUD - Add Route and Controllers for Systems
### Description

we need a route and controller file for the get all systems service function

### Acceptance Criteria

- [ ] new router on endpoint/systems
- [ ] controller to call the service function

### Proposed Solution

look at finishline

epic #8 

### Mocks

_No response_

leoleader:
## Changes

- Added route and controller for getDataByDataType
- Added route and controller for getAllDataTypes
- Added route and controller for getAllSystems
- Added route and controller for getAllRuns
- added the above to index.ts
- added try catches to controllers for proper error handling
- updated NotFoundError in Odyssey-Base from class to function

## Testing
- Manually tested using Postman each route that the request goes through and returns [] since there is nothing in the database rn
- For getDatayDataType, receives 404 response with custom error message
## Screenshots 
Screen Shot 2023-10-09 at 10 22 04 PM
Screen Shot 2023-10-09 at 10 22 37 PM
Screen Shot 2023-10-09 at 10 22 19 PM
Screen Shot 2023-10-09 at 10 22 13 PM
Screen Shot 2023-10-09 at 10 22 26 PM
Screen Shot 2023-10-10 at 9 12 20 PM

## Notes

- Rn checks not successful because need to update submodule Odyssey-Base with changes I made to NotFoundError by merging PR I made (at least I think this is the way to do this i could be wrong)
- Minor Change, in order for the body of req to be read as json and not undefined I needed to make a minor change and move app.use(express.json) higher in the index.ts, before the app.use routes. 

Closes #49, #50, #52


scylla-server/src/controllers/data.controller.ts
```diff
@@ -0,0 +1,16 @@
+import { Request, Response } from 'express';
+import DataService from '../odyssey-base/src/services/data.services';
+
+/**
+ * Controller to manage Data requests and responses.
+ */
+export default class DataController {
+  static async getDataByDataTypeName(req: Request, res: Response) {
```
Peyton-McKee, line 8: you need to add the next: NextFunction param

scylla-server/src/controllers/data.controller.ts
```diff
@@ -0,0 +1,16 @@
...
+    try {
+      const dataByDataTypeName = await DataService.getDataByDataTypeName(req.body.dataTypeName);
+      res.status(200).json(dataByDataTypeName);
+    } catch (error: any) {
+      res.status(error.status).send(error.message);
```
Peyton-McKee, line 13: and then you do if (error instanceof Error) next(error)

Peyton-McKee (changes requested):
nice job on these, just make sure to use the next function instead of the res

Peyton-McKee (approved):
nice job

PR #143, opened 2024-03-17
115 landing dashboard half gauge graph

### LINKED ISSUE #115: Landing Dashboard - Half Gauge graph
### Description

we want a graph to display information with a gauge much like the NERO speed graph

### Acceptance Criteria

- [ ] Configurable height and width
- [ ] takes in a max, min value and then a current value
- [ ] takes in a unit
- [ ] looks like mock 
- [ ] Takes in a color to use for fill

### Proposed Solution

^

### Mocks

Screenshot 2024-02-09 at 10 22 11 AM

leoleader:
UPDATE 3/19
- Switched from NpxGaugeModule to ApexCharts so no new dependency needed
- This implementation is more finnicky, needed to figure out some workarounds: apex radial charts hardcoded for percentages so changed it so calculates value as percentage using given min and max and then labels it w actual val and unit. also ran into trouble with dims and scaling, eventually figured something out that works for size > 100 but still feels a lil iffy. 

apex screenshot
Screenshot 2024-03-19 at 10 33 48 PM

## Changes

- Using NpxGaugeModule https://www.npmjs.com/package/ngx-gauge?activeTab=readme, so need update dependencies
- Half gauge component with customizable colors, size, unit, values, min, max, etc.

## Screenshots
Screenshot 2024-03-16 at 7 56 50 PM
Screenshot 2024-03-16 at 8 20 05 PM

Closes #115

### DIFF
--- a/angular-client/package-lock.json
+++ b/angular-client/package-lock.json
@@ -26,7 +26,7 @@
         "apexcharts": "^3.44.0",
         "mapbox-gl": "^3.1.0",
         "ng-apexcharts": "^1.8.0",
-        "ngx-gauge": "^9.0.0",
+        "ngx-gauge": "^8.0.0",
         "primeng": "^16.9.1",
         "rxjs": "~7.8.0",
         "socket.io-client": "^4.7.2",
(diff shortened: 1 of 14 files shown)


Peyton-McKee (changes requested):
use apex charts instead of ngx gauge

Peyton-McKee (dismissed):
looks great nice work

PR #158, opened 2024-06-13
GHCR Auto built with GH Actions

jr1221:
Changes:
 - Add a workflow which builds scylla and angular for arm64 and amd64, and publishes them to ghcr.io as public, builds upon **manually calling it in the actions tab and PR merged.**.  Rationale: it takes 30+ min to run so only do it when a user manually calls it, and when PR is created and merged.
 - Allows angular-client to be built from the docker image (idk why this wasnt done before?)
 - Edits the compose to be platform agnostic
 - Edits the compose to point to this repository GHCR link

Tested:
- Can run on x86_64, can run on router
- **Did not test with edits done to env**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,8 @@ services:
     restart: unless-stopped
     environment:
       POSTGRES_HOST_AUTH_METHOD: trust
+    ports: # needs to be published for charybdis, which runs outside of docker
+      - 5432:5432
     expose:
       - 5432
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,6 @@ services:
   odyssey-timescale:
     container_name: odyssey-timescale
     image: timescale/timescaledb:2.13.1-pg15
-    platform: linux/arm64
     restart: unless-stopped
     network_mode: host
     environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,6 +11,11 @@ services:
       - 5432:5432
     expose:
       - 5432
+    volumes:
+      - db-data:/var/lib/postgresql/data
+    cpu_shares: 1024
+    mem_limit: 3gb
+    stop_grace_period: 1m
 
   scylla-server:
     container_name: scylla-server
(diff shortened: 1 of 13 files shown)


docker-compose.yml
```diff
@@ -11,6 +11,11 @@ services:
       - 5432:5432
     expose:
       - 5432
+    volumes:
```
RChandler234, line 14: what's the volumes stuff for?

docker-compose.yml
```diff
@@ -47,6 +57,32 @@ services:
...
+    oom_kill_disable: true
+
+    
+
+
+  
+
+
+volumes:
+  db-data:
```
RChandler234, line 80: why does this have to be so far down at the bottom?

RChandler234 (changes requested):

docker-compose.yml
jr1221: The standard specification puts services above volumes as you declare the volumes used by timescale in the service.

docker-compose.yml
jr1221: This allows timescale to persist beyond the lifetime of the router.  Before we just avoided ever properly shutting down service we just pulled the plug.  With this timescale already exists when running docker compose up.  It will save about 2 min boot time (down to about 10 seconds) and will prevent anyone from accidentally losing all data.  I'll add a note in the Readme about this.

docker-compose.yml
jr1221: Here's where I took inspiration.
https://docs.docker.com/compose/compose-file/07-volumes/#example

docker-compose.yml
RChandler234: 🔥

docker-compose.yml
RChandler234: yeah I'm saying why are there like 8 lines between the end of services and this, looks a bit weird

jr1221: Oh lol didn't notice that.  Will shrink it down.

PR #164, opened 2024-07-04
Go rewrite services draft (Jack)

jr1221:


scylla-server-go/services/data_service/data_service.go
```diff
@@ -0,0 +1,50 @@
...
+	"scylla-server/protobuf/serverdata"
+	"scylla-server/services/run_service"
+	"strconv"
+	"time"
+)
+
+func GetDataByDataTypeNameAndRunId(dataTypeName string, runId int) []db.DataModel {
+	client, ctx := prisma.PrismaClient()
+
+	dataType, err := client.DataType.FindUnique(db.DataType.Name.Equals(dataTypeName)).Exec(ctx)
```
Peyton-McKee, line 20: Do you get an error if it doesnt exist?

Peyton-McKee: If so then we want to error out not just console log

scylla-server-go/services/data_service/data_service.go
```diff
@@ -0,0 +1,50 @@
...
+	log.Printf("Err %q\n", err.Error())
```
Peyton-McKee: Will this print regardless if theres an error or not

scylla-server-go/services/data_service/data_service.go
```diff
@@ -0,0 +1,50 @@
...
+
+	var numbers []float64
+	for _, val := range serverData.Values {
+		if n, err := strconv.ParseFloat(val, 64); err == nil {
+			numbers = append(numbers, n)
+		}
+	}
+
+	created, err := client.Data.CreateOne(db.Data.DataType.Link(db.DataType.Name.Set(dataType.Name)), db.Data.Time.Set(time.UnixMilli(unixTime)), db.Data.Run.Link(db.Run.ID.Equals(run.ID)), db.Data.Values.Set(numbers)).Exec(ctx)
```
Peyton-McKee, line 70: We need a formatter

Peyton-McKee: And a linter if we dont have one yet

scylla-server-go/services/node_service/node_service.go
```diff
@@ -0,0 +1,32 @@
...
+	data, err := client.Node.FindMany(db.Node.DataTypes.Every()).Exec(ctx)
+	log.Printf("Err %q\n", err.Error())
+
+	return data
+}
+
+func UpsertNode(nodeName string) *db.NodeModel {
+	client, ctx := prisma.PrismaClient()
+
+	node, err := client.Node.FindUnique(db.Node.Name.Equals(nodeName)).Exec(ctx)
```
Peyton-McKee: Is there no prisma.upsert statement?

scylla-server-go/services/run_service/run_service.go
```diff
@@ -0,0 +1,46 @@
...
+	"scylla-server/prisma/db"
+	"time"
+)
+
+func GetAllRuns() []db.RunModel {
+	client, ctx := prisma.PrismaClient()
+	runs, err := client.Run.FindMany().Exec(ctx)
+	log.Printf("Err %q\n", err.Error())
+
+	// TODO: no mapping done
```
Peyton-McKee: We need mapping from prisma type to shared type

scylla-server-go/services/run_service/run_service.go
```diff
@@ -0,0 +1,46 @@
...
+
+	return runs
+}
+
+func GetRunById(id int) *db.RunModel {
```
Peyton-McKee: These return types should all be shared types...

Peyton-McKee (changes requested):
Overall looks good, looks like what i wouldve expected, def needs some cleaning up though with the types we should be returning and whatnot

scylla-server-go/services/node_service/node_service.go
jr1221: There is but that's not what was used in the JS one.

jr1221:
Fuck go.  Actually go pretty nice its just not as well supported by prisma, and it looks kinda mid.  Plus we get calypso and scylla on the same lang which would be epic.

PR #165, opened 2024-07-05
Rust rewrite: services and basic routing

jr1221:
Rewrite in rust:

Stack
 - Axum for web routing: Reason -> everyone uses it and its got tokio integration
 - Socketoxide for socketio: Reason -> it supports engine v4 and its part of tokio
 - Prisma client rust for ORM: Reason -> its the only one which can take prisma files

This PR:
- Basic CI/CD for devs, inc. format
- all services
- tests for all services
- controllers and transformation functions for all which match client json
- unit tests to ensure client json matches
- routing framework with basic config

Future PR 1:
- Socketoxide
- MQTT connection
- Batch upload in services
- proxy clients
- extended mock support, etc.
- Dockerize, composeify, and document

Future PR 2:
- Further CI/CD fixups
- Axum caching, performance considerations
- Investigate uneeded queries (marked TODO)
- Investigate any structural changes


.github/workflows/scylla-rust-ci.yml
```diff
@@ -0,0 +1,33 @@
+name: Scylla Rust CI
+
+on:
```
Peyton-McKee, line 3: is this the same one as calypso?

scylla-server-rust/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+```
+# in this directory
+SOURCE_DATABASE_URL=postgresql://postgres:password@127.0.0.1:5432/timescaledb cargo run
+```
+
+### Test this app
+
+Since this app uses the database for testing, you must wipe the database.
```
Peyton-McKee: we gotta add scripts for this.

.github/workflows/scylla-rust-ci.yml
jr1221: pretty much.  removed tests bc tests call the db and audit bc we would fail lol.

scylla-server-rust/src/controller/data_controller.rs
```diff
@@ -0,0 +1,42 @@
+use axum::{
+    extract::{Path, State},
+    Json,
+};
+use serde::Serialize;
+
```
RChandler234: rename folder to controllers

scylla-server-rust/src/controller/data_type_controller.rs
```diff
@@ -0,0 +1,29 @@
...
+            name: value.name.clone(),
+            unit: value.unit.clone(),
+        }
+    }
+}
+
+pub async fn get_all_data_types(
+    State(db): State<Database>,
+) -> Result<Json<Vec<DataTypeSend>>, ScyllaError> {
+    let data = data_type_service::get_all_data_types(&db).await?;
```
RChandler234: rename to data_types and transformed_data_types direrctly below

scylla-server-rust/src/controller/data_controller.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+pub async fn get_data(
+    State(db): State<Database>,
+    Path((data_type_name, run_id)): Path<(String, i32)>,
+) -> Result<Json<Vec<DataSend>>, ScyllaError> {
+    let data = data_service::get_data(&db, data_type_name, run_id).await?;
+
+    // map data to correct types
+    let mut transformed_data: Vec<DataSend> = data.iter().map(DataSend::from).collect();
```
RChandler234: I would change this comment to specify what you mean by correct type (type that the frontend uses I'm assuming)

scylla-server-rust/src/controller/data_controller.rs
```diff
@@ -0,0 +1,42 @@
...
+    // sort it by time for the client
+    transformed_data.sort();
```
RChandler234: imo you don't need this comment here cause you can kinda tell by reading what's happening

scylla-server-rust/src/controller/data_controller.rs
```diff
@@ -0,0 +1,42 @@
...
+
+// convert the prisma type to the client type for JSON encoding
+impl From<&prisma::data::Data> for DataSend {
+    fn from(value: &prisma::data::Data) -> Self {
+        DataSend {
+            values: value.values.iter().map(|f| f.to_string()).collect(),
+            time: value.time.timestamp_millis(),
+        }
+    }
+}
```
RChandler234: looks like this method is basically acting as the transformer? I would rephrase the comment to reflect that this is a method with that purpose, but IMO it's fine to keep these in the controller files

scylla-server-rust/src/controller/run_controller.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    let data_new = data.unwrap();
+
+    let transformed_data = RunSend {
+        id: data_new.id,
+        location_name: data_new.location_name.unwrap_or_default(),
+        driver_name: data_new.driver_name.unwrap_or_default(),
+        system_name: data_new.system_name.unwrap_or_default(),
+        time: data_new.time.timestamp_millis(),
+    };
```
RChandler234: why can't you use RunSend::from here?

scylla-server-rust/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+            "/data/:dataTypeName/:runId",
+            get(controller::data_controller::get_data),
+        )
+        .route("/datatypes", get(data_type_controller::get_all_data_types))
+        .route("/drivers", get(driver_controller::get_all_drivers))
+        .route("/locations", get(location_controller::get_all_locations))
+        .route("/nodes", get(node_controller::get_all_nodes))
+        .route("/runs", get(run_controller::get_all_runs))
+        .route("/runs/:id", get(run_controller::get_run_by_id))
+        .route("/systems", get(system_controller::get_all_systems))
```
RChandler234, line 38: It'd be good to separate types of routes with comments, so it's a little easier to read. If we can get this file readable, I think it's fine not to make a bunch of files for 1/2 routes each if that's not very rustlike

scylla-server-rust/src/services/data_service.rs
```diff
@@ -0,0 +1,49 @@
...
+    db.data()
+        .find_many(vec![
+            prisma::data::data_type_name::equals(data_type_name),
+            prisma::data::run_id::equals(run_id),
+        ])
+        .exec()
+        .await
+}
+
+/// Add a datapoint
```
RChandler234: All the service methods should have robust comments specifying parameter types return types, and any weird logic. We use JavaDoc style for FinishLine, looks like RustDoc is the equivalent for Rust

RChandler234: Example of a FinishLine one:
  /**
   * Retrieves a single design review
   *
   * @param submitter the user who is trying to retrieve the design review
   * @param designReviewId the id of the design review to retrieve
   * @param organizationId the organization that the user is currently in
   * @returns the design review
   */

RChandler234 (changes requested):
Haven't worked with Rust a ton, so take what I say with a grain of salt.

.github/workflows/scylla-rust-ci.yml
Peyton-McKee: well were gonna have to add tests back once we do them

scylla-server-rust/src/controller/data_controller.rs
```diff
@@ -0,0 +1,42 @@
...
+
+use crate::{
+    error::ScyllaError,
+    prisma::{self},
+    services::data_service,
+    Database,
+};
+
+#[derive(Serialize, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct DataSend {
```
Peyton-McKee: I would name this PublicData, a typical naming convention for transformed types in servers

scylla-server-rust/src/controller/data_controller.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub time: i64,
+    pub values: Vec<String>,
+}
+
+// convert the prisma type to the client type for JSON encoding
+impl From<&prisma::data::Data> for DataSend {
```
Peyton-McKee: I would move this struct and implementation to a transformers folder, same comments for all the public types

scylla-server-rust/src/controller/data_controller.rs
Peyton-McKee: I would disagree, i think they should be in their own file for nested cases like what if nodes contain data types which contain data, their purpose shouldnt be isolated to the controller

scylla-server-rust/src/controller/data_controller.rs
```diff
@@ -0,0 +1,42 @@
...
+    fn from(value: &prisma::data::Data) -> Self {
+        DataSend {
+            values: value.values.iter().map(|f| f.to_string()).collect(),
+            time: value.time.timestamp_millis(),
+        }
+    }
+}
+
+pub async fn get_data(
+    State(db): State<Database>,
```
Peyton-McKee: why do the controllers need the database? cant the services just get this?

scylla-server-rust/src/controller/data_controller.rs
```diff
@@ -0,0 +1,42 @@
...
+    Path((data_type_name, run_id)): Path<(String, i32)>,
+) -> Result<Json<Vec<DataSend>>, ScyllaError> {
+    let data = data_service::get_data(&db, data_type_name, run_id).await?;
```
Peyton-McKee: if this errors, will it automatically return the Error? Im confused why this is returning a result? this should be a void function that sends the data via whatever http interface we are gonna use. And we should handle errors either from here or with a custom error handler.

scylla-server-rust/src/controller/data_type_controller.rs
```diff
@@ -0,0 +1,29 @@
+use axum::{extract::State, Json};
+use serde::Serialize;
+
+use crate::{error::ScyllaError, prisma, services::data_type_service, Database};
+
+#[derive(Serialize, Debug, PartialEq)]
+pub struct DataTypeSend {
```
Peyton-McKee: same comments as above

scylla-server-rust/src/controller/driver_controller.rs
```diff
@@ -0,0 +1,37 @@
+use axum::{extract::State, Json};
+use serde::Serialize;
+
+use crate::{error::ScyllaError, prisma, services::driver_service, Database};
+
+use super::run_controller::{self, RunSend};
+
+#[derive(Serialize, PartialEq)]
+pub struct DriverSend {
```
Peyton-McKee: same comments as above

scylla-server-rust/src/controller/driver_controller.rs
```diff
@@ -0,0 +1,37 @@
...
+    username: String,
+    runs: Vec<run_controller::RunSend>,
```
Peyton-McKee: see this shouldnt be getting inmproted from the controller the object should exist outside of its respective controller.

scylla-server-rust/src/controller/location_controller.rs
```diff
@@ -0,0 +1,43 @@
+use axum::{extract::State, Json};
+use serde::Serialize;
+
+use crate::{error::ScyllaError, prisma, services::location_service, Database};
+
+use super::run_controller::{self, RunSend};
+
+#[derive(Serialize, Debug, PartialEq)]
+pub struct LocationSend {
```
Peyton-McKee: same comments as above

scylla-server-rust/src/controller/node_controller.rs
```diff
@@ -0,0 +1,36 @@
+use axum::{extract::State, Json};
+use serde::Serialize;
+
+use crate::{error::ScyllaError, prisma, services::node_service, Database};
+
+use super::data_type_controller::DataTypeSend;
+
+#[derive(Serialize, PartialEq)]
+pub struct NodeSend {
```
Peyton-McKee: same comments as above

scylla-server-rust/src/controller/run_controller.rs
```diff
@@ -0,0 +1,62 @@
+use axum::{
+    extract::{Path, State},
+    Json,
+};
+use serde::Serialize;
+
+use crate::{error::ScyllaError, prisma, services::run_service, Database};
+
+#[derive(Serialize, Debug, PartialEq)]
+pub struct RunSend {
```
Peyton-McKee: same coments as above

scylla-server-rust/src/controller/run_controller.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+pub async fn get_run_by_id(
+    State(db): State<Database>,
+    Path(run_id): Path<i32>,
+) -> Result<Json<RunSend>, ScyllaError> {
+    let data = run_service::get_run_by_id(&db, run_id).await?;
+
+    if data.is_none() {
+        return Err(ScyllaError::NotFound);
```
Peyton-McKee: this shouldnt be in the controller, we should be erroring inside the service

scylla-server-rust/src/controller/system_controller.rs
```diff
@@ -0,0 +1,37 @@
+use axum::{extract::State, Json};
+use serde::Serialize;
+
+use crate::{error::ScyllaError, prisma, services::system_service, Database};
+
+use super::run_controller::{self, RunSend};
+
+#[derive(Serialize, Debug, PartialEq)]
+pub struct SystemSend {
```
Peyton-McKee: same comments as above

scylla-server-rust/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+pub mod services;
+
+#[allow(clippy::all)]
+#[allow(warnings)]
+pub mod prisma;
+
+pub mod serverdata;
+
+/// The type descriptor of the database passed to the middlelayer through axum state
+pub type Database = std::sync::Arc<prisma::PrismaClient>;
```
Peyton-McKee, line 13: i would call this prisma, I mean I get calling it database, however its just the prisma client, which weve called just prisma in other applications. not a big deal, i was a little confused when you said we were able to pass around a reference to the database, thats technically not what this is doing but I get why you might call it that

scylla-server-rust/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+    prisma::PrismaClient,
+    Database,
+};
+use tower_http::cors::{Any, CorsLayer};
+
+#[tokio::main]
+async fn main() {
+    let client: Database = Arc::new(PrismaClient::_builder().build().await.unwrap());
+
+    let app = Router::new()
```
Peyton-McKee, line 18: we should have different routers for all of our types, most of them just have one route rn, but in the interest of scalability we should be building it more modulely

scylla-server-rust/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+        .route("/runs/:id", get(run_controller::get_run_by_id))
+        .route("/systems", get(system_controller::get_all_systems))
+        .layer(
+            CorsLayer::new()
+                // allow `GET`
+                .allow_methods([Method::GET])
+                // allow requests from any origin
+                .allow_origin(Any),
+        )
+        .with_state(client);
```
Peyton-McKee, line 46: is this the suggested way of doing this? is it not possible to just have a global we just import into each of the service functions, I would get it if rust doesnt like that though. However it makes me uncomfy to have the prisma client inside the controllers when they really dont need it

scylla-server-rust/src/services/data_service.rs
```diff
@@ -0,0 +1,49 @@
...
+    run_id: i32,
+) -> Result<prisma::data::Data, QueryError> {
+    db.data()
+        .create(
+            prisma::data_type::name::equals(data_type_name),
+            DateTime::from_timestamp_millis(unix_time)
+                .unwrap()
+                .fixed_offset(),
+            prisma::run::id::equals(run_id),
+            vec![prisma::data::values::set(
```
Peyton-McKee, line 49: very interesting syntax

scylla-server-rust/tests/data_service_test.rs
```diff
@@ -0,0 +1,128 @@
...
+use test_utils::cleanup_and_prepare;
+
+const DATA_TYPE_NAME: &str = "test";
+
+#[tokio::test]
+async fn test_data_service() -> Result<(), QueryError> {
+    let db = cleanup_and_prepare().await?;
+
+    run_service::create_run_with_id(&db, 0, 0).await?;
+    node_service::upsert_node(&db, DATA_TYPE_NAME.to_owned()).await?;
```
Peyton-McKee: creating a node with a data type name?

scylla-server-rust/tests/data_service_test.rs
```diff
@@ -0,0 +1,128 @@
...
+    data_type_service::upsert_data_type(
+        &db,
+        DATA_TYPE_NAME.to_owned(),
+        "joe_mama".to_owned(),
+        DATA_TYPE_NAME.to_owned(),
+    )
+    .await?;
+    data_service::get_data(&db, DATA_TYPE_NAME.to_owned(), 0).await?;
+
+    Ok(())
```
Peyton-McKee, line 30: theres no asserts here?

Peyton-McKee: or is this just a smoke test?

scylla-server-rust/tests/data_type_service_test.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    // should fail since no node exists
+    data_type_service::upsert_data_type(
+        &db,
+        DATA_TYPE_NAME.to_owned(),
+        "hello wurld".to_owned(),
+        DATA_TYPE_NAME.to_owned(),
+    )
+    .await
+    .expect_err("Should error, as no node exists");
```
Peyton-McKee: we should be expecting the proper error that its gonna have

Peyton-McKee (changes requested):
looks great

scylla-server-rust/src/services/data_service.rs
jr1221: rustdoc doesnt support any annotations such as params, so it wont parse them or anything nice :( Added in latest commit.

scylla-server-rust/src/controller/data_controller.rs
jr1221: the axum/tower/tokio way of doing things is a very seperate state for the lifetime of each route.  Each route is given a context and state (the db is in the Arc so it can contain references across threads) and if it leaves its state anything it modifies can be poisoned without a mutex.  By adhering to this rule, we can ensure we never need to worry about mutexing or RwLocking anything, as axum handles the threads and load balancing for us.

scylla-server-rust/src/controller/data_controller.rs
jr1221: Axum has two error handlers, of which I implemented the infallible one so far.  This basically means a controller cannot and will not panic.  If it does panic in the controller it moves on to the fallible error handler, which i have not done yet.  Basically returning a result means the Ok() value is the payload sent to the client and error is sent to the error handler system in `error.rs`.  The `IntoResponse` is called therefore processing the error to send to the client.  This has multiple benefits, including allowing for a very effortless `?` operator for any result we have to `unwrap()` usually.   It also allows for custom error handling to be statically typed, which increases readability and safety assurances.

scylla-server-rust/src/controller/run_controller.rs
jr1221: I disagree on this, as I think the service doesnt care about the lack of existence of data, in my opinion it should just return what the database fetches.  Doing this in the controller allows the http-related stuff to be propogated independently of the service, so therefore unit tests and other callers could use the service functions without needing to handle http-related helpers.  Sorry I am probably over the top on this but I really like the idea of really simple and stupid services that only adhere to the prisma library standards.

scylla-server-rust/src/lib.rs
jr1221: is `PrismaDatabase` or something more unique alright?  The namespace gets hella confusing as the library is called `prisma` and the inner type is called `PrismaClient`.

scylla-server-rust/src/main.rs
jr1221: See my above comment on passing `Database` in.  Its related to async and tokio stuff.  I get you are worried about it being where it doesnt need to be, but I cant think of a different way thats also thread safe, load balancing safe, etc.

scylla-server-rust/tests/data_service_test.rs
jr1221: oh thats just me being lazy not wanted to use the string "test".  Ill fix that.

scylla-server-rust/tests/data_service_test.rs
jr1221: its a "doesnt return err test", how it was originally written I believe.

scylla-server-rust/src/lib.rs
Peyton-McKee: eh we can leave it as Database then.

scylla-server-rust/tests/data_service_test.rs
Peyton-McKee: yeah thats called a smoke test

scylla-server-rust/src/main.rs
jr1221: Since the app is ultimately one namespace with one state, one CORS, etc. I really would hesitate to create a second router.  I do not know the implications of creating multiple CORS and multiple state `Database` etc, but I worry that it will lead to increase memory usage and bugs.  

When would a second router become beneficial for example in the express scylla?

This is the closest to a way to do that, but it doesnt mention state so Idk what will happen there: https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.merge

I guess I could also make a bunch of nested services.

scylla-server-rust/tests/data_type_service_test.rs
jr1221: Unfortunately the prisma client rust query errors are not typed.  It basically is just a wrapper for the error the database returned as a string.  Kinda sucks I know.

PR #173, opened 2024-08-01
Scylla Calypso

jr1221:
## Changes

- Adds an endpoint to send a car configuration.  This is a POST request to `config/<key>?data=123&data=456`.  For context, this sends a MQTT message in accordance with the specifications of [this PR](https://github.com/Northeastern-Electric-Racing/Calypso/pull/44).
- Adds new error types and reorganizes some code to get the MQTT publisher into the controller's scope
- Adds the new protobuf specification for the data sent to Calypso
- Reorganizes some of the processing code to make it ignore command messages for decoding, and also cleanup up a little issue with batch upload looping

## Notes

Do yall have any recommendations on how error codes and whatnot should be handled in the web server side?  Should an error message be returned in the body of the response, or should we just have error codes and a log of what happened internal to the server?

## Test Cases

- Working with local Calypso on the bidirectionality branch.  Tested by seeing the calypso feedback and via `candump` ing a virtual can interface

## Checklist

- [oops] All commits are tagged with the ticket number
- [X] No linting errors / newline at end of file warnings
- [X] All code follows repository-configured prettier formatting
- [X] No merge conflicts
- [X] All checks passing
- [X] Screenshots of UI changes (see Screenshots section)
- [X] Remove any non-applicable sections of this template
- [X] Assign the PR to yourself
- [X] No `package-lock.json` changes (unless dependencies have changed)
- [lol] PR is linked to the ticket (fill in the closes line below)

### DIFF
--- a/scylla-server-rust/src/controllers/car_command_controller.rs
+++ b/scylla-server-rust/src/controllers/car_command_controller.rs
@@ -17,19 +17,20 @@ pub struct ConfigRequest {
     pub data: Option<Vec<f32>>,
 }
 
-/// This controller recieves two peices of info: the key from the path, and the list of float values from the query params
-/// The key is used in accordance with the calypso specification: a command_data object is sent over siren to CALYPSO_BIDIR_CMD_PREFIX/<key>
-/// That command_data object is processed and matched via the key to a YAML specification for the encoding and sending of a CAN message.
-/// If the data is invalid, too short, etc. the default feild from the YAML is instead sent, as with when the car initially turns on before argos connects.
-/// TLDR: This triggers a command <key> with the values <data_query.0.data> for calypso to update the CAN packet sent to the bus.
-pub async fn send_config(
+/// Sends a configuration to the car over MQTT
+/// * `key` - The key of the configuration, as defined in the cangen YAML
+/// * `data_query` - The data of the configuration, a URI query list of data=<f32>.  If empty or too short, filled with cangen YAMl defaults
+/// * `client` - The MQTT client to be used to send the data
+///
+/// More info: This follows the specification of sending a command_data object over siren to topic CALYPSO_BIDIR_CMD_PREFIX/<key>
+pub async fn send_config_command(
     Path(key): Path<String>,
-    data_query: Query<ConfigRequest>,
+    Query(data_query): Query<ConfigRequest>,
     Extension(client): Extension<Option<Arc<AsyncClient>>>,
 ) -> Result<(), ScyllaError> {
     info!(
         "Sending car config with key: {}, and values: {:?}",
-        key, data_query.0.data
+        key, data_query.data
     );
     // disable scylla if not prod, as there will be None mqtt client
     let Some(client) = client else {
@@ -38,14 +39,14 @@ pub async fn send_config(
         ));
     };
 
-    // create a payload object of the values to be parsed by calypso into a CAN packet
+    // the protobuf calypso converts into CAN
     let mut payload = CommandData::new();
-    // writing an empty protobuf is OK, that indicates to calypso that the default value should be used
-    if let Some(data) = data_query.0.data {
+    // empty "data" in the protobuf tells calypso to use the default value
+    if let Some(data) = data_query.data {
         payload.data = data;
     }
     let Ok(bytes) = payload.write_to_bytes() else {
-        return Err(ScyllaError::ImpossibleEncoding(
+        return Err(ScyllaError::InvalidEncoding(
             "Payload could not be written!".to_string(),
         ));
     };
--- a/scylla-server-rust/src/controllers/car_command_controller.rs
+++ b/scylla-server-rust/src/controllers/car_command_controller.rs
@@ -34,9 +34,8 @@ pub async fn send_config_command(
     );
     // disable scylla if not prod, as there will be None mqtt client
     let Some(client) = client else {
-        return Err(ScyllaError::NotProd(
-            "Car config sending is disabled in mock mode!".to_string(),
-        ));
+        warn!("Cannot use config endpoint in dev mode!");
+        return Ok(())
     };
 
     // the protobuf calypso converts into CAN
--- a/scylla-server-rust/src/controllers/car_command_controller.rs
+++ b/scylla-server-rust/src/controllers/car_command_controller.rs
@@ -9,6 +9,7 @@ use tracing::{info, warn};
 
 use crate::{command_data::CommandData, error::ScyllaError};
 
+/// the prefix for the calypso topic, so topic of cmd is this plus the key appended on
 pub const CALYPSO_BIDIR_CMD_PREFIX: &str = "Calypso/Bidir/Command/";
 
 #[derive(Deserialize, Debug)]
@@ -16,6 +17,10 @@ pub struct ConfigRequest {
     pub data: Vec<f32>,
 }
 
+/// This controller recieves two peices of info: the key from the path, and the list of float values from the query params
+/// The key is used in accordance with the calypso specification: a command_data object is sent over siren to CALYPSO_BIDIR_CMD_PREFIX/<key>
+/// That command_data object is processed and matched via the key to a YAML specification for the encoding and sending of a CAN message
+/// TLDR: This triggers a command <key> with the values <data_query.0.data> for calypso to update the CAN packet sent to the bus.
 pub async fn send_config(
     Path(key): Path<String>,
     data_query: Query<ConfigRequest>,
@@ -25,16 +30,19 @@ pub async fn send_config(
         "Sending car config with key: {}, and values: {:?}",
         key, data_query.0.data
     );
+    // disable scylla if not prod, as there will be None mqtt client
     let Some(client) = client else {
         return Err(ScyllaError::NotProd);
     };
 
+    // create a payload object of the values to be parsed by calypso into a CAN packet
     let mut payload = CommandData::new();
     payload.data = data_query.0.data;
     let Ok(bytes) = payload.write_to_bytes() else {
         return Err(ScyllaError::ImpossibleEncoding);
     };
 
+    // publish the message to the topic that calypso's encoder is susbcribed to
     if let Err(err) = client
         .publish(
             format!("{}{}", CALYPSO_BIDIR_CMD_PREFIX, key),
@@ -47,5 +55,6 @@ pub async fn send_config(
         warn!("Could not publish instruction: {}", err);
         return Err(ScyllaError::CommFailure);
     }
+
     Ok(())
 }
--- a/scylla-server-rust/src/controllers/car_command_controller.rs
+++ b/scylla-server-rust/src/controllers/car_command_controller.rs
@@ -14,12 +14,13 @@ pub const CALYPSO_BIDIR_CMD_PREFIX: &str = "Calypso/Bidir/Command/";
 
 #[derive(Deserialize, Debug)]
 pub struct ConfigRequest {
-    pub data: Vec<f32>,
+    pub data: Option<Vec<f32>>,
 }
 
 /// This controller recieves two peices of info: the key from the path, and the list of float values from the query params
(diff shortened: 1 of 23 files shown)


scylla-server-rust/src/controllers/mod.rs
```diff
@@ -4,4 +4,5 @@ pub mod driver_controller;
 pub mod location_controller;
 pub mod node_controller;
 pub mod run_controller;
+pub mod send_config_controller;
```
RChandler234: So the naming for this is kind of weird cause this controller works differently from the rest of them since it sends commands to the car instead of receiving them. I'd name this car_command_controller or config_command_controller. We want the names to be object oriented instead of action oriented if that makes sense

scylla-server-rust/src/processors/db_handler.rs
```diff
@@ -112,30 +112,26 @@ impl DbHandler {
...
-                    } else {
-                     info!("No more messages to cleanup.");
-                     break;
+                    // cleanup all remaining messages if batches start backing up
+                    while let Some(final_msgs) = batch_queue.recv().await {
+                        info!("{} batches remaining!", batch_queue.len()+1);
+                    info!(
+                        "A cleanup batch uploaded: {:?}",
+                        data_service::add_many(&database, final_msgs).await
+                    );
```
RChandler234, line 129: what's this code doing?

scylla-server-rust/src/main.rs
```diff
@@ -186,6 +198,11 @@ async fn main() {
         )
         // SYSTEMS
         .route("/systems", get(system_controller::get_all_systems))
+        // CONFIG
+        .route(
+            "/config/:key",
+            post(send_config_controller::send_config).layer(Extension(client)),
```
RChandler234: can you rename this, 'config/set/:key', it just makes the intention that this is supposed to a post a little more clear.

Also, IMO key should be changed to configName, or configKey if key is the name we're using everywhere else

scylla-server-rust/src/controllers/send_config_controller.rs
```diff
@@ -0,0 +1,51 @@
...
+use crate::{command_data::CommandData, error::ScyllaError};
+
+pub const CALYPSO_BIDIR_CMD_PREFIX: &str = "Calypso/Bidir/Command/";
+
+#[derive(Deserialize, Debug)]
+pub struct ConfigRequest {
+    pub data: Vec<f32>,
+}
+
+pub async fn send_config(
```
RChandler234: can you add a detailed comment to this/ throw some comments throughout this method, since it's probably gonna be used for a bunch of different commands/ it's not immediately obvious to me what all the code is doing. 

Also, IMO, we should just make different endpoints for each of the different configs that someone would want to set. That way, we'll have a lot more control over what range of values are acceptable and maybe even some kind of permissions system if we want to gate some configs behind a password or smthn

RChandler234 (changes requested):

scylla-server-rust/src/processors/db_handler.rs
jr1221: That code batch uploads all remaining batches before exiting, such that scylla will not exit until all batch uploads are complete.

basically I noticed I used a _loop_ and the proper rust syntax is a _while_ and it annoyed me so much I switched it in this PR even though this PR is unrelated. my b.

scylla-server-rust/src/controllers/car_command_controller.rs
```diff
@@ -0,0 +1,70 @@
...
+#[derive(Deserialize, Debug)]
+pub struct ConfigRequest {
+    pub data: Option<Vec<f32>>,
+}
+
+/// This controller recieves two peices of info: the key from the path, and the list of float values from the query params
+/// The key is used in accordance with the calypso specification: a command_data object is sent over siren to CALYPSO_BIDIR_CMD_PREFIX/<key>
+/// That command_data object is processed and matched via the key to a YAML specification for the encoding and sending of a CAN message.
+/// If the data is invalid, too short, etc. the default feild from the YAML is instead sent, as with when the car initially turns on before argos connects.
+/// TLDR: This triggers a command <key> with the values <data_query.0.data> for calypso to update the CAN packet sent to the bus.
```
RChandler234: Why are you sending the default field if the data is invalid? Shouldn't it just reject it and send an error back to scylla?

RChandler234: Also if you need to have a tl;dr on a comment, you should probably reword it a bit. I'd format it in some formal kind of way to describe params, errors, returns, etc. it makes it easier to read. I know java has a standard format for its comments, not sure if there's one for rust

scylla-server-rust/src/controllers/car_command_controller.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// the prefix for the calypso topic, so topic of cmd is this plus the key appended on
+pub const CALYPSO_BIDIR_CMD_PREFIX: &str = "Calypso/Bidir/Command/";
+
+#[derive(Deserialize, Debug)]
+pub struct ConfigRequest {
+    pub data: Option<Vec<f32>>,
+}
+
+/// This controller recieves two peices of info: the key from the path, and the list of float values from the query params
```
RChandler234: some spelling errors on this line

scylla-server-rust/src/controllers/car_command_controller.rs
```diff
@@ -0,0 +1,70 @@
...
+/// The key is used in accordance with the calypso specification: a command_data object is sent over siren to CALYPSO_BIDIR_CMD_PREFIX/<key>
+/// That command_data object is processed and matched via the key to a YAML specification for the encoding and sending of a CAN message.
+/// If the data is invalid, too short, etc. the default feild from the YAML is instead sent, as with when the car initially turns on before argos connects.
+/// TLDR: This triggers a command <key> with the values <data_query.0.data> for calypso to update the CAN packet sent to the bus.
+pub async fn send_config(
```
RChandler234: send_config_command*

scylla-server-rust/src/controllers/car_command_controller.rs
```diff
@@ -0,0 +1,70 @@
...
+    Extension(client): Extension<Option<Arc<AsyncClient>>>,
+) -> Result<(), ScyllaError> {
+    info!(
+        "Sending car config with key: {}, and values: {:?}",
+        key, data_query.0.data
+    );
+    // disable scylla if not prod, as there will be None mqtt client
+    let Some(client) = client else {
+        return Err(ScyllaError::NotProd(
+            "Car config sending is disabled in mock mode!".to_string(),
```
RChandler234: imo, just log a message and return if we're in development mode, no need to error

scylla-server-rust/src/controllers/car_command_controller.rs
```diff
@@ -0,0 +1,70 @@
...
+        ));
+    };
+
+    // create a payload object of the values to be parsed by calypso into a CAN packet
+    let mut payload = CommandData::new();
+    // writing an empty protobuf is OK, that indicates to calypso that the default value should be used
```
RChandler234: Try to make these comments a little more concise. The info's good, but the more efficient you can be with the number of words, the better

scylla-server-rust/src/controllers/car_command_controller.rs
```diff
@@ -0,0 +1,70 @@
...
+    if let Some(data) = data_query.0.data {
+        payload.data = data;
+    }
+    let Ok(bytes) = payload.write_to_bytes() else {
+        return Err(ScyllaError::ImpossibleEncoding(
```
RChandler234: I'd call this error "InvalidEncoding"

scylla-server-rust/src/controllers/run_controller.rs
```diff
@@ -25,7 +25,7 @@ pub async fn get_run_by_id(
     let run_data = run_service::get_run_by_id(&db, run_id).await?;
 
     if run_data.is_none() {
-        return Err(ScyllaError::NotFound);
+        return Err(ScyllaError::EmptyResult);
```
RChandler234, line 28: why the name change?

scylla-server-rust/src/error.rs
```diff
@@ -10,30 +10,49 @@ use tracing::warn;
...
     PrismaError(QueryError),
-    NotFound,
+    /// Not available in mock mode, which the system is in
+    NotProd(String),
+    /// An instruction was not encodable
+    ImpossibleEncoding(String),
+    /// Could not communicate to car
+    CommFailure(String),
+    /// A query turned up empty that should not have
+    EmptyResult,
```
RChandler234, line 19: put comments to the right instead of above each enum name, looks cleaner

scylla-server-rust/src/processors/mqtt_processor.rs
```diff
@@ -161,14 +164,23 @@ impl MqttProcessor {
...
     #[instrument(skip(self), level = Level::TRACE)]
     fn parse_msg(&self, msg: Publish) -> Result<ClientData, impl fmt::Debug> {
-        let data = serverdata::ServerData::parse_from_bytes(&msg.payload)
-            .map_err(|f| format!("Could not parse message topic:{:?} error: {}", msg.topic, f))?;
+        let topic = std::str::from_utf8(&msg.topic)
+            .map_err(|f| format!("Could not parse topic: {}, topic: {:?}", f, msg.topic))?;
+
+        // ignore command messages, less confusing in logs than just failing to decode protobuf
+        if topic.starts_with(CALYPSO_BIDIR_CMD_PREFIX) {
+            return Err(format!("Skipping command message: {}", topic));
```
RChandler234, line 172: wait I'm confused, why would scylla ever get a command message?

RChandler234 (changes requested):

scylla-server-rust/src/controllers/car_command_controller.rs
jr1221: Ok will do.  Invalid is the wrong term I used.  If the data is the wrong number of points or is null, then the default is sent.  This is a way to "reset to default" the config setting.

scylla-server-rust/src/controllers/car_command_controller.rs
jr1221: Ok.  What does the client recieve back in that case?

scylla-server-rust/src/error.rs
jr1221: Alright.  I don't think rust docs will pick up the comment however.

scylla-server-rust/src/controllers/run_controller.rs
jr1221: because not found is an error I used for when a query cannot find something.  On other other hand this error should not be that, instead it should represent when runs are found but the result is empty.

scylla-server-rust/src/processors/mqtt_processor.rs
jr1221: Because the MQTT server sends to everyone subscribed to the topic the command.  And scylla is subscribed to every topic so it can store them in it's database.

scylla-server-rust/src/error.rs
jr1221: Ya in fact rust will still count the comment as the item below the comment leading it to be confusing.

RChandler234: Ugh, that is dumb as hell, makes sense tho

scylla-server-rust/src/processors/mqtt_processor.rs
RChandler234: Ahhh ok, makes sense

scylla-server-rust/src/controllers/car_command_controller.rs
RChandler234: I was thinking just nothing, we shouldn't have any need to test sending commands to the car in dev mode/ if we do I can set up some way to mock it, but I'd rather any buttons or anything we hook up right now just do nothing locally

PR #180, opened 2024-08-12
#178 - Graph Resizing

### LINKED ISSUE #178: Fix Graph Resizing
### Description

Currently all the graphs resize in a weird way where it shifts all the points on the graph. Due to a lot of requests from engineering/ the fact that I suspect this way of re-rendering is more cumbersome to re-render, we want to try to make the graph slide along as it updates, or at least re-render in a way that looks less weird/ more performant

### Acceptance Criteria

- [ ] New way of graph resizing/ updating should look better/ be more performant
- [ ] Any other existing functionality should be unaffected

### Proposed Solution

Look into Apex Charts, see if there's another way to live update the graphs

### Mocks

N/A

RChandler234:
## Notes

I added in a set range for the graph (currently less than one minute - any longer and the wobbling would occur before the graph reached the end of the range). This seems to consistently solve the issue until right around when the graph reaches the left side of the screen, then it reverts back to the old way sporadically.
Still investigating why that's happening, I have a hunch that either something is causing something to render too slowly which causes the issue to start happening or maybe multiple data points are getting added for the same point in time (from the graph's pov) [I'm less convinced of that one because then why is it only happening once the graph makes it to the left side of the range].
- edit: ok yeah upon further investigation, about a minute in is when it starts fucking up on develop as well, so I think it's some issue with how we're updating the data being passed to the graph

Anyways, will look into that, just wanted to make a draft PR for documenting purposes.

N.B. It seems like only the most recent minute (give or take) of data is being saved. Mostly just documenting this for myself

This article was pretty helpful to explain what actually causes the wiggle. I'd still like to try out D3 sometime in the future, just to see if I could get it any more performant (also just I'm interested), but pretty hopeful that I can get something consistently working with Apex for now: https://bost.ocks.org/mike/path/

## Checklist

- [ ] All commits are tagged with the ticket number
- [ ] No linting errors / newline at end of file warnings
- [ ] All code follows repository-configured prettier formatting
- [ ] No merge conflicts
- [ ] All checks passing
- [ ] Screenshots of UI changes (see Screenshots section)
- [ ] Remove any non-applicable sections of this template
- [ ] Assign the PR to yourself
- [ ] No `package-lock.json` changes (unless dependencies have changed)
- [ ] PR is linked to the ticket (fill in the closes line below)

Closes #178  (issue #)

### DIFF
--- a/angular-client/src/components/acceleration-graphs/acceleration-graphs.component.ts
+++ b/angular-client/src/components/acceleration-graphs/acceleration-graphs.component.ts
@@ -41,15 +41,6 @@ export class AccelerationGraphs implements OnInit {
         y: y1
       });
 
-      //limits the data storage to 400 to prevent lag
-      if (this.xData.length > this.maxDataPoints) {
-        this.xData = this.xData.slice(1);
-      }
-
-      if (this.yData.length > this.maxDataPoints) {
-        this.yData = this.yData.slice(1);
-      }
-
       //checks if there is a new max
       this.xMax = Math.max(Math.abs(x1), this.xMax);
       this.yMax = Math.max(Math.abs(y1), this.yMax);
--- a/angular-client/src/components/acceleration-over-time-display/acceleration-over-time-display.component.ts
+++ b/angular-client/src/components/acceleration-over-time-display/acceleration-over-time-display.component.ts
@@ -10,16 +10,12 @@ import { GraphData } from 'src/utils/types.utils';
 })
 export default class AccelerationOverTimeDisplay {
   data: GraphData[] = [];
-  maxDataPoints = 100;
 
   constructor(private storage: Storage) {}
 
   ngOnInit() {
     this.storage.get(IdentifierDataType.ACCELERATION).subscribe((value) => {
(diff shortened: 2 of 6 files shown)


RChandler234:
The flow graph works,  the kicker is that you can only add points to the graph or it'll do the wobbling thing. This is because when you pop a value off the beginning of the array of graph values, Apex is registering it as a whole new set of points and thinks it has to interpolate between the old points and the new points, even though they're all mostly the same. 

I honestly think this is kinda cool, cause if we add in a "pause graph" button, it means you could go back and look at chunks of data on a graph in realtime like you do with historical data.

My only concern with this is that if you leave the realtime graph running for too long, it might lag the page because the graph is storing too many values (even though it's only displaying a certain range), but it hasn't caused an issue with limited testing, so I'm optimistic that Apex is doing something in the background to avoid this issue (like not rendering values at all if they're outside the currently displayed range). 

Will do more extensive testing/ comment the result (want to apply this same change to the graphs on the landing page if there's no real hit to the render-time)

jr1221:
Bug: does not clear graph when switching datapoints.

PR #185, opened 2024-09-13
build(webapp/deps): migrate to Angular v18

Char2sGu:


angular-client/src/index.html
```diff
@@ -8,6 +8,6 @@
     <link href="https://api.tiles.mapbox.com/mapbox-gl-js/v1.4.1/mapbox-gl.css" rel="stylesheet" />
   </head>
   <body>
-    <app-context />
+    <app-context></app-context>
```
Char2sGu, line 11: Custom Elements are not allowed to be self-closed. `<app-context />` was an invalid syntax, and after migration, this causes the build to fail

angular-client/src/app/app.module.ts
```diff
@@ -190,19 +191,21 @@ import { BatteryInfoDisplay } from 'src/pages/charging-page/components/battery-i
...
-  bootstrap: [AppContext]
+  providers: [
+    DialogService,
+    MessageService,
+    provideHttpClient(withInterceptorsFromDi())
+  ]
 })
 export class AppModule {
-  constructor(
+  constructor (
```
Char2sGu: This formatting issue will be addressed later by forcing formatting and linting in Git Hooks

angular-client/src/app/app.module.ts
```diff
@@ -190,19 +191,21 @@ import { BatteryInfoDisplay } from 'src/pages/charging-page/components/battery-i
...
     MatButtonModule,
     MatInputModule,
     ReactiveFormsModule
   ],
-  providers: [DialogService, MessageService],
-  bootstrap: [AppContext]
+  providers: [
+    DialogService,
+    MessageService,
+    provideHttpClient(withInterceptorsFromDi())
```
Char2sGu: Angular is going "standalone". The `HttpClientModule` is deprecated and replaced by the standalone provider factory `provideHttpClient`

Char2sGu: The new http interceptor api uses functional interceptors. In order to enable class-based interceptors, `withInterceptorsFromDi` is necessary.

RChandler234 (changes requested):
This fucked with the sidebar design a bit somehow. I honestly like the new one better, but could you make it so the item highlight doesn't overflow on the right side
New:
image
Old:
image

Char2sGu:
Sure. Since we bumped the version of Angular Material and PrimeNG, some updates in these packages might result in unexpected behaviors. I’ll look into that today.

Char2sGu:
image

RChandler234 (approved):
LGTM 🔥

PR #208, opened 2024-09-20
initial static rate limit

### LINKED ISSUE #183: [Scylla] - Implement basic static rate limiting
### Description

Scylla JS has rate limiting.  Turns out we need it for rust as well.  It basically involves discarding any message that is sent within 100 ms (by default) of the last message per topic name.  

### Acceptance Criteria

- Can change the overall minimum time
- If done after #182, make this an atomic setting with an edit endpoint
- Performance hit is minimal

### Proposed Solution

At the db handler layer introduce a HashMap<String, DateTime> or something which stores the time the last message came in.  Do this only for messages that are not considered special types (i.e. messages we plan to batch insert).  If the difference in time of a new message is smaller than the 100ms or whatever setting, discard the message and bail out of the processor.  If it is greater store the new time in the hashmap and continue processing the data.

### Mocks

_No response_

jr1221:
Basic rate limit.

100ms --> 4403 messages
0ms --> 17643 messages

This is messy and could be improved on.  Need to revise function signatures and cleanup processor.

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -32,8 +32,8 @@ services:
         #-SCYLLA_DATA_UPLOAD_DISABLE=false
         #-SCYLLA_SIREN_HOST_URL=localhost:1883
         #-SCYLLA_BATCH_UPSERT_TIME=10
-        #-SCYLLA_RATE_LIMIT_MODE=none
-        #-SCYLLA_STATIC_RATE_LIMIT_VALUE=100
+        - SCYLLA_RATE_LIMIT_MODE=static
+        - SCYLLA_STATIC_RATE_LIMIT_VALUE=50
         #-SCYLLA_SOCKET_DISCARD_PERCENT=0
         - RUST_LOG=warn,scylla_server=debug
     cpu_shares: 1024
--- a/compose.yml
+++ b/compose.yml
@@ -28,6 +28,13 @@ services:
         - SOURCE_DATABASE_URL=postgresql://postgres:password@odyssey-timescale:5432/postgres
        # - PROD_SIREN_HOST_URL=siren:1883
         - SCYLLA_PROD=true
+        #- SCYLLA_SATURATE_BATCH=false
+        #-SCYLLA_DATA_UPLOAD_DISABLE=false
+        #-SCYLLA_SIREN_HOST_URL=localhost:1883
+        #-SCYLLA_BATCH_UPSERT_TIME=10
+        #-SCYLLA_RATE_LIMIT_MODE=none
+        #-SCYLLA_STATIC_RATE_LIMIT_VALUE=100
+        #-SCYLLA_SOCKET_DISCARD_PERCENT=0
         - RUST_LOG=warn,scylla_server=debug
     cpu_shares: 1024
     stop_grace_period: 2m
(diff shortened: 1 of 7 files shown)


jr1221:
Also fixes scylla ID.

compose.yml
```diff
@@ -28,6 +28,13 @@ services:
         - SOURCE_DATABASE_URL=postgresql://postgres:password@odyssey-timescale:5432/postgres
        # - PROD_SIREN_HOST_URL=siren:1883
         - SCYLLA_PROD=true
+        #- SCYLLA_SATURATE_BATCH=false
+        #-SCYLLA_DATA_UPLOAD_DISABLE=false
+        #-SCYLLA_SIREN_HOST_URL=localhost:1883
+        #-SCYLLA_BATCH_UPSERT_TIME=10
+        - SCYLLA_RATE_LIMIT_MODE=static
+        - SCYLLA_STATIC_RATE_LIMIT_VALUE=50
+        #-SCYLLA_SOCKET_DISCARD_PERCENT=0
```
RChandler234, line 37: why all this commented out?

scylla-server/src/processors/mqtt_processor.rs
```diff
@@ -9,50 +8,72 @@ use rumqttc::v5::{
...
 use crate::{
     controllers::car_command_controller::CALYPSO_BIDIR_CMD_PREFIX, serverdata,
-    services::run_service,
+    services::run_service, RateLimitMode,
 };
 
 use super::ClientData;
 use std::borrow::Cow;
 
+/// a mqtt processor
```
RChandler234: if you're gonna have this here, you could describe what it's doing more

scylla-server/src/processors/mqtt_processor.rs
```diff
@@ -66,16 +87,21 @@ impl MqttProcessor {
             .set_session_expiry_interval(Some(u32::MAX))
             .set_topic_alias_max(Some(600));
 
+        let rate_map: HashMap<String, Instant> = HashMap::new();
+
         // TODO mess with incoming message cap if db, etc. cannot keep up
```
RChandler234: this basically what you're doing rn right?

scylla-server/src/processors/mqtt_processor.rs
```diff
@@ -170,26 +203,41 @@ impl MqttProcessor {
...
         if topic.starts_with(CALYPSO_BIDIR_CMD_PREFIX) {
-            return Err(format!("Skipping command message: {}", topic));
+            debug!("Skipping command message: {}", topic);
+            return None;
         }
 
-        let split = topic
-            .split_once('/')
-            .ok_or(&format!("Could not parse nesting: {:?}", msg.topic))?;
+        if self.rate_limit_mode == RateLimitMode::Static {
```
RChandler234, line 224: Put a comment here explaining what this code is doing/ why. The code is a bit complicated, and we definitely want somebody else to understand what's going on/ why down the line

RChandler234 (changes requested):

compose.yml
jr1221: this is just a way for me to document the settings without opening the code.

RChandler234: heard, sounds good

scylla-server/src/processors/mqtt_processor.rs
jr1221: ya.

RChandler234: go ahead and remove the todo unless you still need it for some reason

PR #252, opened 2024-11-29
223 metadata injector

### LINKED ISSUE #207: More sustainable way to store timer values (for charging screen timers)
### Description

Currently total times are stored in as values in sessionStorage, to allow for the total time to be saved if the page reloads. However, these timers are not true to the entire charging session.  We need a way to have the timers that track charging status's that is actually true. That is either first messages received by the client or preferably values tracked using the server.

### Acceptance Criteria

Timers for charging status's (connected to the combined-status-display component) are true to values read from the car.

### Proposed Solution

Talk to Reid or Wyatt (Wyatt might need to talk to Wyatt).

### Mocks

_No response_

### LINKED ISSUE #223: [Scylla] - Add Scylla Metadata Injection
### Description

Right now, all data addition or expansion is done strictly by the client.  Some transformation is done by Calypso but it is very limited in scope.  Therefore issues like #207 will continue to be a problem and the end users won't like this. 

### Acceptance Criteria

This would be the addition of a new thread to Scylla which intercepts and manipulates socket messages before they are sent.  This would strictly add new messages or deny others, it should not mutate existing messages. Extensibility for any sort of data manipulation is key, as well as organization and little performance overhead in the mqtt_processor thread.

Requirements for merge:
- does not reduce performance of mqtt processor or cause it to idle busy
- Adds a timer method for things charging needs (consult @bracyw)
- does not rely on unabstracted code for such timers
- Allows to pass through all messages which do not have special hooks very quickly. 
- Does not mutate existing data, just adds more

### Proposed Solution

The best way in my view to add this thread is to split the socket from the mqtt processor.  Then, switch the mpsc channel to be broadcast type so two such consumer exist, the db Handler and the metadata injector.  This metadata injector would then receive every message and manipulate, add messages, etc.  It could use tokio select! patterns as seen in other threads to also run periodic tasks.  This metadata counting would be hard coded, which is ugly but it's OK.

Depends on #change 

### Mocks

_No response_

jr1221:
## Changes

Adds a metadata injector.  It basically splits off socket functionality from mqtt_processor into its own file, and hides the metadata behind a flag.

Currently, the `timers`  socketIO key:
```
{
  "topic": "BMS/Charging/Control",
  "last_change": "2025-01-18T02:40:09.521458544Z",
  "last_value": 0
}
```

Currently, the `faults` socketIO key (sample):
```
[
  {
    "node": "Bms",
    "name": "DCL_Enforce",
    "occured_at": "2025-01-18T02:39:36.972566Z",
    "last_seen": "2025-01-18T02:40:09.506620Z",
    "expired": false
  },
  {
    "node": "Bms",
    "name": "External_Can",
    "occured_at": "2025-01-18T02:39:36.972585Z",
    "last_seen": "2025-01-18T02:40:09.506653Z",
    "expired": false
  },
  {
    "node": "Bms",
    "name": "Charge_Reading",
    "occured_at": "2025-01-18T02:39:36.972731Z",
    "last_seen": "2025-01-18T02:40:09.506750Z",
    "expired": false
  },
]
```

Currently

## Notes

- The fault and timer re-use and buffering logic is shaky and uses some assumptions: A fault that does not broadcast in 8+ seconds is considered a new fault.  A timer cannot be a fault. There will never be more than 25 concurrent faults (lol).
- Depends on https://github.com/Northeastern-Electric-Racing/Embedded-Base/issues/225 for MPU faulting to be detected.

## To Do

- [ ] Figure out what to do about latency sender

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #223
Closes #207


scylla-server/src/socket_handler.rs
```diff
@@ -0,0 +1,285 @@
...
+}
+
+#[derive(Serialize, Clone)]
+struct FaultData {
+    /// the node the fault came from
+    pub node: Node,
+    /// the word describing the fault
+    pub name: String,
+    /// when the fault occured
+    #[serde(with = "ts_milliseconds")]
```
bracyw: why are these both ts milliseconds when serialized

scylla-server/src/socket_handler.rs
```diff
@@ -0,0 +1,285 @@
...
+                } else {
+                    continue;
+                };
+
+                trace!("Matched on {}, {:?}", flt_txt, node);
+
+                // default to sending a new fault
+                let mut should_push = true;
+                // iterate through current faults
+                for item in fault_ringbuffer.iter_mut() {
```
bracyw: If there have been no fault's yet wouldn't it push whatever data we get because the for loop never iterates? ... or is the fault_ringbuffer never empty?

scylla-server/src/socket_handler.rs
```diff
@@ -0,0 +1,285 @@
...
+                send_socket_msg(
+                    &data,
+                    &mut upload_counter,
+                    upload_ratio,
+                    &io,
+                    DATA_SOCKET_KEY,
+                );
+
+                // check to see if we fit a timer case, and then act upon it
+                // IMPORTANT: assumes a timer is never also a fault
```
bracyw: if we ever wanted to track timer on faults we could probably just add that in the fault mqtt logic. Or have a different enum for fault timers. Gonna write this down cause certainly something we could use for the fault page.

scylla-server/src/socket_handler.rs
```diff
@@ -0,0 +1,285 @@
...
+                break;
+            },
+            Ok(data) = data_channel.recv() => {
+                send_socket_msg(&data, &mut upload_counter, upload_ratio, &io, DATA_SOCKET_KEY);
+            }
+        }
+    }
+}
+#[derive(Serialize)]
+struct TimerData {
```
bracyw: maybe structs for this should be put in different file so we have all data stuff for sockets together.

bracyw (commented):
Looks good. minor questions we can talk about tonight if your pulling up. Didn't test yet on my end.

scylla-server/src/socket_handler.rs
```diff
@@ -0,0 +1,285 @@
...
+
+    loop {
+        tokio::select! {
+            _ = cancel_token.cancelled() => {
+                debug!("Shutting down socket handler!");
+                break;
+            },
+            Ok(data) = data_channel.recv() => {
+                send_socket_msg(
+                    &data,
```
bracyw, line 80: idk if rust will get made with the async and borrow bs... but if this could be split into helper methods. That would make adding functionality + readability so much better. Idk what your time looks like but it would probably be good to get this merged and then I can make some small cleanup tickets (so we don't get into conflict hell)

scylla-server/src/socket_handler.rs
jr1221: so the client receives them as milliseconds since unix epoch

scylla-server/src/socket_handler.rs
jr1221: Removing the continue on line 155 should help

scylla-server/src/socket_handler.rs
jr1221: Yes.  But whatever data we get is already a fault, as above we continue to next iteration of loop if the data isnt a fault.

bracyw (approved):
lgtm - test using postman

Issue #243, opened 2024-11-20
[Scylla] - Investigate a new CSV handler


jr1221:
### Description

Investigate a new CSV handler that when enabled would store data into a CSV file(s) that can then be uploaded with Charybdis.

### Acceptance Criteria

- New runtime enabled csv handler
- Compatable with charybids, uplaods correctly
- Any change to schema is somehow reflected in csv handler or csv handler fails to compile (maybe use diesel types?)

### Proposed Solution

- Adds a setting in main.rs to switch to csv handler, option for csv path, spawn csv handler or db_handler depending on mode select
- Add a csv_handler.rs which takes data and writes to CSV using the diesel types possibly as a guide
- Pipe data from mqtt_processor to csv_handler instead of DB handler
- Write a charybdis script that can handle that style CSVs in one simple command
- Benchmark performance and whatnot
- Test on cloud, fetch data with Grafana

jr1221:
New idea:
- A logger on odysseus serializes the MQTT to a new protobuf format
- A script on odysseus sends the files to router scylla
- Scylla processes the files and upserts as necessary
- Then normal charybdis uploads the whole local database on the router

PR #262, opened 2024-12-21
Diesel Performance Fixes, Batching Improvements, New Allocator

### LINKED ISSUE #184: [Scylla] - Investigate performance improvements
### Description

Now that the rust implementation is finished, there could be many performance improvements added to it.  These boil down to three categories.  CPU time, memory allocation, and polling/async frequencies.

### Acceptance Criteria

- Async: Investigate polling/async frequencies and ensure `select!` is not providing an overhead, probably using `tokio_console`
- Memory alloc: Investigate any data cloning in hot routes, like the query or especially the batch insert, and eliminate them.
- CPU time: Investigate the use of rayon to parallelize db batching or other large functions.
- CPU Time: Search for any computational overheads and see if SIMD compliant functions exist.

### Proposed Solution

Possibly rearrange selects into dedicated tasks, implement rayon `par_iter`, implement simd common, etc.

### Mocks

_No response_

### LINKED ISSUE #244: [Scylla] - Fix batching logic
### Description

Batching logic has a max instruction set of 65535, and with 4 instructions per INSERT the batch has a max messages of 64435/4.  We need to optimize splitting the batch inserts while continuing to obey the upsert time.

### Acceptance Criteria

- New logic splits up inserts so there isnt any 1 item inserts

### Proposed Solution

Split the items up using a division, where remainder is attached to one message, guarranteeing data over the limit wont be used.

### Mocks

_No response_

jr1221:
## Changes

- Switch to diesel-async, to better use the tokio runtime to organize threads
Rationale:  Performance of blocking threads decreases with more threads in use, affecting large chunk batching and the #243 code
- Switch to jemalloc, as extreme memory use was being observed with #243 
Rationale: The system allocator was not freeing the memory used by the batching tasks effectively, resulting in excessive memory usage.  It takes now approximately 1-2gb of RAM to upload an hour of data, and that RAM is released back when the upload is complete.  Previously it was 5-10gb, and the memory was not freed.
- Modify the values field of the DATA table to be NOT NULL, modify the values field to be REAL instead of DOUBLE. PRECISION.
Rationale (found in the docs for DataInsert): This aligns closer to the actual data we recieve, which is not null 4 byte floating point integers.  This allows less re-allocation of data when converting our recieved protobuf to the data the database is looking for, improving performance.
- Redo chunking logic. 
Rationale:  Before, chunks of data to upload were split into a few even chunks and one chunk of only a few points.  The new algorithm better evens out the chunks.  It still could use improvement, and more investigation into the libpq instruction limit could be warrented.

## Notes

Jemalloc may fail to build on aarch64, hence the CI running, so we should pay attention to any unsoundness or stability issues.

## Test Cases

- All functionality normal, perhaps a little faster.
## To Do

- [x] Ensure CI works

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #244
Closes #184


scylla-server/src/db_handler.rs
```diff
@@ -12,19 +13,46 @@ use crate::{ClientData, PoolHandle, RUN_ID};
 /// upserting of metadata for data, and batch uploading the database
 pub struct DbHandler {
     /// The list of data types seen by this instance, used for when to upsert
-    datatype_list: Vec<String>,
+    datatype_list: FxHashSet<String>,
```
bracyw, line 16: Why the switch to FxHashSet?

scylla-server/src/db_handler.rs
```diff
@@ -81,17 +110,16 @@ impl DbHandler {
...
                         continue;
                     }
-                    debug!("Batch uploading {} chunks in parrallel", msgs.len() / chunk_size);
-                    for chunk in msgs.chunks(chunk_size).collect::<Vec<_>>() {
-                        let owned = chunk.to_vec();
-                        let pool = pool.clone();
-                       tokio::task::spawn_blocking(move || {
-                            DbHandler::batch_upload(owned, pool)});
+                    let msg_len = msgs.len();
+                    let chunk_size = msg_len / ((msg_len / 8190) + 1);
```
bracyw, line 118: Why are you now dividing by 8190 instead of 16380, something special about dividing max params by 8 instead of 4?

scylla-server/src/db_handler.rs
```diff
@@ -143,17 +171,31 @@ impl DbHandler {
         data_channel: Sender<Vec<ClientData>>,
         cancel_token: CancellationToken,
     ) {
+        let mut batch_interval = tokio::time::interval(Duration::from_millis(self.upload_interval));
+        // the match batch size to reasonably expect
```
bracyw: `*` the max

scylla-server/src/db_handler.rs
```diff
@@ -143,17 +171,31 @@ impl DbHandler {
...
                     break;
                 },
                 Some(msg) = self.receiver.recv() => {
                     self.handle_msg(msg, &data_channel).await;
                 }
+                _ = batch_interval.tick() => {
+                    if !self.data_queue.is_empty() {
+                        // set a new max if this batch is larger
+                        max_batch_size = usize::max(max_batch_size, self.data_queue.len());
+                        // mem::take allows us to assign the value of data queue vec::new() while maintaining the memory for data_channel ownership
```
bracyw: where is mem::take ?

scylla-server/src/db_handler.rs
```diff
@@ -81,17 +110,16 @@ impl DbHandler {
                     // libpq has max 65535 params, therefore batch
```
bracyw, line 110: I actually never understood this fully. why is the batching logic needed for diesel... is it because it let's you try and insert more data then libpq can handle. Does prisma just manage this itself?

bracyw (changes requested):
Integration tests fail sometimes when you either have the database disabled in docker or haven't set it up yet. I think just adding a sleep like your github tests. For now we can probably just restart or remove the database if it is already a docker container but disabled.... I can push changes for this if you want.

everything compiles tests and runs fine on mac m1.

jr1221:
Ya can unadd the fix to integration tests?  Thanks.

scylla-server/src/db_handler.rs
jr1221: 1. A hashset should be used whenever you are checking for uniqueness rather than storing possibly duplicate items.
2. Rust has multiple hashing algos builtin, the default is slow but secure, since we don't need security we can switch to a faster one.  This library fxhash is just a syntactic sugar for switching the hashing algorithm. For example the official rust compiler uses fxhash instead of the default.  It is considerably faster and considered best practice.  I measured about .25% speed time reduction in the overall time usage hashing in the whole program when I switched, which is considerable.

scylla-server/src/db_handler.rs
jr1221: Oh ya forgot to investigate that.  The switch to diesel async doubled the number of instructions per insert.  I'll do some investigating there. It's a very annoying limit.

scylla-server/src/db_handler.rs
jr1221: Oops will remove.  Mem take is pretty cool tho, but turns out I didn't need it.

scylla-server/src/db_handler.rs
jr1221: I guess prisma splits the queries as needed.  I doubt they work around libpq as it's kinda the best way to communicate with postgres.  It's a very annoying limit but kinda inherent to postgres.

bracyw (approved):
LGTM

PR #280, opened 2025-02-09
Create dependabot.yml

jr1221:
Automatic dependency PR updates sent out once a week.  See calypso for good example of use.  Good to keep up with perf fixes and other updates.


jr1221:
Can also prob due node if ur interested.
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#example-dependabotyml-file

PR #312, opened 2025-03-03
Expose Settings from Argos

### LINKED ISSUE #182: Expose certain settings over the webserver for Argos setting
### Description

Some settings should be admin/user facing on Argos.  Provide a HTTP post endpoint for each of them with param.

### Acceptance Criteria

- Can toggle SCYLLA_DATA_UPLOAD_DISABLE
- Can set SCYLLA_SOCKET_DISCARD_PERCENT
- Can set SCYLLA_BATCH_UPSERT_TIME
- Can toggle SCYLLA_SATURATE_BATCH 

### Proposed Solution

Use Atomic* types declared above main() in order to provide the functionality.  

Have a POST endpoint for each setting (`argos/<setting>/set`) using a query or param (idk best) with the relevant Atomic passed in as a extension with a static lifetime.  The controller will call `load` with relaxed ordering.

Change the functions that use these settings to take in an atomic type, and check the Atomic* type at each action.  Ensure performance regressions are not major with the Atomic* types by using relaxed ordering on store and load.

### Mocks

_No response_

jr1221:
## Changes

Makes the following settings global via an atomic variable, and provides put set requests and an overall get request.

- Data upload disable
- batch time
- ratelimit mode
- ratelimit static time
- socket discard percentage

## Notes

This can then be used for an Argos GUI to set the above settings.

Please review the naming of the endpoints, settings, and overall API bc Idk what is best settable by client.

## Test Cases

- Changed all settings observed effect, also observed the get_settings update

## Checklist

- [X] All code follows repository-configured prettier formatting
- [X] No merge conflicts
- [x] All checks passing

Closes #182

### DIFF
--- a/scylla-server/src/db_handler.rs
+++ b/scylla-server/src/db_handler.rs
@@ -1,3 +1,5 @@
+use std::sync::atomic::Ordering;
+
 use rustc_hash::FxHashSet;
 use tokio::sync::{broadcast, mpsc};
 
@@ -7,7 +9,7 @@ use tokio_util::sync::CancellationToken;
 use tracing::{debug, info, instrument, trace, warn, Level};
 
 use crate::services::{data_service, data_type_service};
-use crate::{ClientData, PoolHandle};
+use crate::{ClientData, PoolHandle, DATA_UPLOAD_DISABLE};
 
 /// A few threads to manage the processing and inserting of special types,
 /// upserting of metadata for data, and batch uploading the database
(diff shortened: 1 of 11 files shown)


scylla-server/src/db_handler.rs
```diff
@@ -56,99 +56,89 @@ fn chunk_vec<T: Clone>(input: Vec<T>, max_chunk_size: usize) -> Vec<Vec<T>> {
...
-                    let Ok(mut database) = pool.get().await else {
-                        warn!("Could not get connection for cleanup");
+            if DATA_UPLOAD_DISABLE.load(Ordering::Relaxed) {
+                tokio::select! {
+                    _ = cancel_token.cancelled() => {
+                        warn!("Cancelling fake upload with {} batches left in queue!", batch_queue.len());
+                        break;
+                    },
+                    Some(msgs) = batch_queue.recv() => {
+                        warn!("NOT UPLOADING {} MESSAGES", msgs.len());
```
bracyw: maybe storing instead of uploading... because uploading kinda feels like cloud or pushing stuff to the client. I'll edit that.

scylla-server/src/main.rs
```diff
@@ -283,11 +272,42 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             "/runs/update/{id}/{driver}/{location}/{notes}",
             post(run_controller::update_run_with_data),
         )
-        // CONFIG
+        // CAR CONFIG
         .route(
             "/config/set/{configKey}",
             post(car_command_controller::send_config_command).layer(Extension(client_sharable)),
         )
+        // SCYLLA CONFIG
```
bracyw, line 280: Going to make a ticket to make different routers files for each parent endpoint. dw about it on this ticket just commenting to remind myself.

scylla-server/src/mqtt_processor.rs
```diff
@@ -171,11 +162,18 @@ impl MqttProcessor {
         }
 
         // handle static rate limiting mode
-        if self.rate_limit_mode == RateLimitMode::Static {
+        if RATE_LIMIT_MODE
```
bracyw, line 165: Some of these additions seem to add some overhead and complexity. Just want to make sure performance and quality is worth the addition. You know better than me about the usefulness of allowing for dynamic settings, so just wanted to double check it is worth it for each settable value.

scylla-server/src/db_handler.rs
```diff
@@ -171,10 +161,20 @@ impl DbHandler {
...
         cancel_token: CancellationToken,
     ) {
-        let mut batch_interval = tokio::time::interval(Duration::from_millis(self.upload_interval));
+        let mut batch_interval = tokio::time::interval(Duration::from_secs(
+            BATCH_UPSERT_TIME.load(Ordering::Relaxed).into(),
+        ));
         // the max batch size to reasonably expect
         let mut max_batch_size = 2usize;
         loop {
+            // if the batch interval changed, act. this is run per message, which is a performance concern
```
bracyw, line 170: If loading the atomic is itself is not a performance concern but rather just when the value changes, it's not concerning. If we did want to avoid this, we can just couple changing the upsert_time with restarting the loop. Because users probably won't/shouldn't be changing this often, so we could simply create a new handler or batching loop when we change static or upsert_time. Just an option lmk what you think.

scylla-server/src/db_handler.rs
```diff
@@ -171,10 +161,20 @@ impl DbHandler {
...
+            if BATCH_UPSERT_TIME.load(Ordering::Relaxed) as u64 != batch_interval.period().as_secs()
```
bracyw, line 171: load upsert once into local variable to use for comparison and new interval.

bracyw (commented):
Looks good couple questions on strategy for handling changes to atomics.

scylla-server/src/mqtt_processor.rs
jr1221: its pretty much the only way to do it, the reasoning here is is it more overhead to use a Mutex versus a Duration::from ?  I assumed a mutex was more overhead due to the negotiation, but i wasnt sure and still arent.

scylla-server/src/db_handler.rs
jr1221: the overhead is in the right side of the equals sign actually, the .load is minimal.

bracyw: All good, I was just thinking that the variable u assign inside the if could be different then when u check it. But I don't have a good reason why that would be a problem for this.

scylla-server/src/mqtt_processor.rs
bracyw: I was thinking more that we require a new service (init a new dbhandler with different settings) to be started whenever a user changes the setting to avoid having global variables. Although probably to complex anyways.

bracyw (approved):
Lgtm

PR #350, opened 2025-03-23
Added Pie Chart

KarimJC:
Working pie chart with updating values. No animation on the pie chart anymore.

## Changes

The pie chart now correctly displays values for the energy usage from the motor controller, battbox fans, LV boards, and pumps. The animation does not exist for the pie chart anymore.


compose/compose.calypso.yml
```diff
@@ -1,7 +1,7 @@
 services:
     calypso:
         container_name: calypso
-        image: ghcr.io/northeastern-electric-racing/calypso:develop
```
bracyw: This can now just be switched back to what it was originally: `ghcr.io/northeastern-electric-racing/calypso:develop`

angular-client/src/components/pie-chart/pie-chart.component.ts
```diff
@@ -65,6 +68,10 @@ export default class PieChartComponent implements OnInit {
       labels,
       legend: {
         offsetX: 10
+      },
+      animations: {
```
bracyw: Check and see if this is still necessary

angular-client/src/components/motor-info/motor-info.component.ts
```diff
@@ -15,22 +15,60 @@ export default class MotorInfoComponent implements OnInit {
...
+      this.updatePieChart();
+    })
+    this.storage.get(DataTypeEnum.PUMPS).subscribe((value) => {
+      this.pumps = floatPipe(value.values[0]);
+      this.updatePieChart();
+    })
+    this.storage.get(DataTypeEnum.MOTOR_CONTROLLER).subscribe((value) => {
+      this.motorController = floatPipe(value.values[0]);
+      this.updatePieChart();
+      console.log(this.piechartData);
```
bracyw: remove console.log

angular-client/src/components/motor-info/motor-info.component.ts
```diff
@@ -15,22 +15,60 @@ export default class MotorInfoComponent implements OnInit {
   motorUsage: number = 100;
   coolUsage: number = 0;
   motorTemp: number = 0;
+  motorController: number = 25;
```
bracyw: Can you make these all out of 20 (so if each was at 5 we would be at 100% usage). And make a 5th value (label: 'None', color grey)  that is anything left over (e.g. if all of these variables are at 4 (16 total) then the 5th value would be 4 to add the total up to 20). lmk if that makes sense.

bracyw (changes requested):
Looking good couple minor touch ups and you'll be finished! Besides the comments below, can you also add back the Screenshots and Checklist sections, just copy paste from another pr. And read through checklist, also add screenshots of new design once you finished changes.  <- that part can be done on Tuesday, most important is actually making code changes.

also to fix prettier run: `npx prettier --write .` to apply prettier format in `angular-client` directory.

PR #432, opened 2025-09-12
#431 Fixed Starting Values for Charging Timers  

### LINKED ISSUE #431: [Charging Page] - Charging Timers Broken on Production
### Observed Behavior

On production the main value for a our Charging Status timers sometimes don't update and say their start time was @ UNIX epoch (0 ms), which displays as a giant number on the frontend. The total time a status has been active is also almost always displays as zero on the frontend, but this is a backend issue because the backend doesn't actually populate the total_timer_map (keyword search for this to find backend logic)

### Expected Behavior

Each timer accurately displays the amount of time that a given status has been "active". The frontend logic for detecting "active" status is correct. As well as the total time it has been "active". 

### Steps to Reproduce

1. Meet with Wyatt for more context.
2. Connect to live backend in the bay (while the car is charging)
3. Use postman to check socket data for the "timers" value
4. see that total_timer_map is empty, and some last_time_values are 0 (should be large ms value)

### Screenshots (as needed)

Image
example ^

bracyw:
## Changes

The timer values would error when the car was already charging on start, because the starting values were either not changed (because the starting value was zero, just like the initialized value), or/and the starting time was 0 and was never updated because of that.

TLDR: fixes charging timers when not on constantly changing mock data

## Notes

Also added mock data to scylla-dev and formatted with prettier.

The time starting at "now" may not be needed. Since the first value for the timer will always be different and should always change the time. Will figure out after testing next time the car charges.

NEEDS TOO BE TESTED NEXT TIME THE CAR IS BEING CHARGED. (going to build scylla latest of this branch.)

 

## Test Cases

TBD, needs to work when car is charging and scylla is turned on after 

## Checklist

- [ ] All commits are tagged with the ticket number
- [ ] No linting errors / newline at end of file warnings
- [ ] All code follows repository-configured prettier formatting
- [ ] No merge conflicts
- [ ] All checks passing
- [ ] Screenshots of UI changes (see Screenshots section)
- [ ] Remove any non-applicable sections of this template
- [ ] Assign the PR to yourself
- [ ] No `package-lock.json` changes (unless dependencies have changed)
- [ ] PR is linked to the ticket (fill in the closes line below)

Closes #431

### DIFF
--- a/compose/compose.scylla-dev.yml
+++ b/compose/compose.scylla-dev.yml
@@ -1,19 +1,25 @@
- services:    
-    scylla-server: # do not run scylla
-        entrypoint: ["echo", "DISABLED"]
-        restart: "no"
-        
+services:
+  scylla-server: # do not run scylla
+    entrypoint: ["echo", "DISABLED"]
+    restart: "no"
 
-    client:
-      extends:
-        file: ../angular-client/compose.client.yml
-        service: client
-      build:
-        context: ../angular-client
-      environment:
-        BACKEND_URL: http://192.168.100.11:8000
+  client:
+    extends:
+      file: ../angular-client/compose.client.yml
+      service: client
+    build:
+      context: ../angular-client
+    environment:
+      BACKEND_URL: http://127.0.0.1:8000
 
-    siren:
-        extends:
-            file: ../siren-base/compose.siren.yml
-            service: siren
+  siren:
+    extends:
+      file: ../siren-base/compose.siren.yml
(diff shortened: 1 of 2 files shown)


compose/compose.scylla-dev.yml
```diff
@@ -1,19 +1,25 @@
- services:
```
bracyw, line 1: this is prettier, which our compose files should be formatted with anyways?

PR #459, opened 2025-10-28
Added functions for mean downsampling with and without raw sql queries

### LINKED ISSUE #457: [Graph Page] - Downsampling for time based queries
### Description

There is a current implementation of downsampling by run id (returning less every nth data point when there is too much data for a run). We now need to abstract that logic and apply it too the get data by time endpoint.

see for this ticket implementation for impl of run id verison: 
https://github.com/Northeastern-Electric-Racing/Argos/pull/441 

### Acceptance Criteria

- Downsampling is abstracted (to a reasonable level) OR change downsampling to be a database level thing
- Run downsampling still works
- time based downsampling works

### Proposed Solution

- Investigate downsampling at database level for diesel.rs and then weigh pro's/con's (might have to do raw sql for this approach)
- If downsampling is to scuffed or complicated, abstract logic of downsampling at the service level.
- Check frontend to make sure everything works, and data is still displayed.

important note: Client changes should not be required, just like the other PR for runs.

### Mocks

_No response_

suryatho:
## Changes

Added capabilities for mean downsampling with raw sql query when getting data by run id

## Notes

Left code for tracing timing of get downsampled data call in data_service.rs

It seems like the data returned by the versions of the method that doesn't use a raw sql query and the one that does don't differ other than usually just one timestamp is different. The data returned by n-th point downsampling vs mean downsampling is different (oh really?). Looks like the mean downsampling does work from eye test, but I haven't written any tests for it. 

I am unsure of how to gather timing data because there is differences when I call the function the first time versus subsequent times, and even then it seems to fluctuate after the first call, so all the different methods (n-th vs. mean vs. mean with raw sql) didn't really show any differences to me. How should I actually test my changes?

## Test Cases

- Didn't add tests yet

## Screenshots

image

## To Do

- [x] Bench changes

## Checklist

- [ ] No merge conflicts
- [x] All checks passing

Closes #457


suryatho:
Should the `get_data_by_run_id_with_auto_downsampling()` function also return the downsampling info in the [`data_controller`](https://github.com/Northeastern-Electric-Racing/Argos/blob/2aac071b6b3b1fc86cd8492504809d5e932280cc/scylla-server/src/controllers/data_controller.rs#L21-L67), or would that be weird ?

bracyw:
> Should the `get_data_by_run_id_with_auto_downsampling()` function also return the downsampling info in the [`data_controller`](https://github.com/Northeastern-Electric-Racing/Argos/blob/2aac071b6b3b1fc86cd8492504809d5e932280cc/scylla-server/src/controllers/data_controller.rs#L21-L67), or would that be weird ?

I think to answer both the question about timing and the question about whether or not to still use  `get_data_by_run_id_with_auto_downsampling()` in the data controller (if that's what your asking), I would say disable the info level print statements (do this by commenting out the tracing subscriber in main.rs), and then just print the time the query started and ended.

An easier probably better way would be just write two tests that loop over the service functions and collect the time taken from invocation of the function to return time, and take the average of like 1,000 calls or something. Database should have a lot of data, run the sim for like 15 minutes and then stop calypso and siren, before running your test. (If you want to you could also setup and impl testing data (like seed data for the integration_test.sh script)

That was a lot of yapping so just @ me in slack with any questions that come up from this.

suryatho:
I just added a file that runs some benchmarks on the two different mean downsampling methods. Its a sad day but the output lowk makes sense. I also tried to analyze the sql query itself but I didn't really know what to do with the data from that.

Screenshot 2025-11-03 at 12 06 05 AM

As you can see the normal implementation just using diesel and rust is significantly faster though we probably aren't seeing any of it on the front end. 
Just a note because I didn't include it in the out, there was only around 8k points for the segment temp data and 10k points for the accelerometer data (1 point per vs 3 points per) so I just fed in a higher sampling rate than it would normally do (it wouldn't even try downsampling in these cases). Maybe that affects the data enough to see in the benchmark but I don't really know if it would.

Oh yeah let me know what I should remove from this branch before it goes back into develop

bracyw:
> I just added a file that runs some benchmarks on the two different mean downsampling methods. Its a sad day but the output lowk makes sense. I also tried to analyze the sql query itself but I didn't really know what to do with the data from that.
> 
> Screenshot 2025-11-03 at 12 06 05 AM
> As you can see the normal implementation just using diesel and rust is significantly faster though we probably aren't seeing any of it on the front end. Just a note because I didn't include it in the out, there was only around 8k points for the segment temp data and 10k points for the accelerometer data (1 point per vs 3 points per) so I just fed in a higher sampling rate than it would normally do (it wouldn't even try downsampling in these cases). Maybe that affects the data enough to see in the benchmark but I don't really know if it would.
> 
> Oh yeah let me know what I should remove from this branch before it goes back into develop

It seems like the first test shows a preference towards raw SQL and the second one is very strong on diesel. Was the mock data shutoff while you run, also I would recommend maybe running on a larger sample set the 10k. Just because the two tests are so different in timing. If you try a larger sample size and it's still very different, then we can just go with diesel. Just want to make sure everything is benchmarked evenly as we can before making a decision.

suryatho:
I added to the bench script to do a bench for each separate data type and write to a file so I can look at the data ([here](https://docs.google.com/spreadsheets/d/1VmgILM2M2Lfm6iOgVRqBt-gN0zmG10ATRtAGXCnzMUI/edit?usp=sharing)). I settled on 30 iterations for each data type with 5 warm-up runs (I don't know if that actually does anything), with 30 the whole benchmarking operation took 30 minutes and I didn't want to run it again. At the top of the spreadsheet I noted some important statistics:

image

It feels like it's contradicting itself at first, but what I think this means is that for more data points the raw SQL query handles it quicker, but for smaller ones, the diesel implementation works better. Also, based on the standard deviation data the SQL query seems to be more consistent but I also don't really know statistics and I don't know if how I am calculating standard deviation or interpreting it is even right.

Inside the code, there are currently two separate implementations of the `get_mean_downsampled_data()` in `data_service.rs`, so one should probably go. Let me know what else should be removed such as new dependencies.

scylla-server/src/services/data_service.rs
```diff
@@ -82,35 +85,136 @@ pub async fn add_many(
...
+        }
+
+        let sum_time: i128 = chunk.iter().map(|d| d.time as i128).sum();
+        let mean_time: i64 = (sum_time / chunk.len() as i128) as i64;
+
+        let min_values_len = chunk.iter().map(|d| d.values.len()).min().unwrap_or(0);
+        let mut sum_values: Vec<f32> = vec![0.0; min_values_len];
+        for d in chunk {
+            for (i, v) in d.values.iter().take(min_values_len).enumerate() {
+                if let Some(val) = v {
```
bracyw: We should really never get a None version of the value here but if we do we should just print a warning message, or error message using the trace subscriber setup. (you can look at batching_loop in db_handler for an example)

scylla-server/src/services/data_service.rs
```diff
@@ -82,35 +85,136 @@ pub async fn add_many(
...
+    }
+
+    Ok(out)
+}
 
-    // Simple downsampling: keep every Nth point
-    // Expand to other downsampling algorithms later
-    let downsampled: Vec<Data> = all_data
+#[derive(QueryableByName)]
+struct AggRow {
```
bracyw: I know we talked about but you can delete the raw sql stuff

bracyw (changes requested):
I know we talked about on Thursday but thought I would just put changes needed to merge now.

bracyw (approved):

PR #477, opened 2025-11-17
Add BiMultiMap to `rule_structs.rs` and use it in RuleManager Move

### LINKED ISSUE #475: Transition rules manager to rule to client relationship
### Description

Currently the rules manager uses a client to rule structure that requires O(all_clients) iterations for each MQTT message:
```rust
clients_map: FxHashMap<ClientId, FxHashMap<Topic, Vec<Rule>>>,
rules_lookup: FxHashMap<Topic, FxHashSet<ClientId>>,
```

> The current timing is not the worst as our client subscription size would most likely be at a max of 10 for the foreseeable future, however, logic-wise this is helpful and ultimately good to support early on.

**Transition to:**
```rust
pub struct RuleManager {
    rules: FxHashMap<RuleId, Rule>,
    topic_index: RwLock<FxHashMap<Topic, FxHashSet<RuleId>>>,
    subscriptions: RwLock<BiMultiMap<RuleId, ClientId>>,
}
```

**BiMultiMap** provides O(1) bidirectional lookups:
- Rule → Clients (for MQTT notifications)
- Client → Rules (for UI focused queries (what rules am I subbed too?)

> Ultimately this will be transitioned to a database in the future, which should be seem less as the rules manager acts as a wrapper for whatever logic we decide it needs.

### Acceptance Criteria

- [ ] functionality stays the same (test with Postman before changes)
- [ ] MQTT message processing uses O(1) topic lookup
- [ ] all operations stay within O(n) time
- [ ] concurrent MQTT processing enabled

### Proposed Solution

1. Create `BiMultiMap<L, R>` struct (ask @bracyw (me) for more details)
2. Create new RuleManager structure with the new fields above
3. Convert old functions to logic for the new structure

suryatho:
- RwLocks into RuleManager for more granular control of blocking
- Implemented borrow for Topic struct so we strings can lend their ownership to topic when all we really need is &Topic

## Changes

See #475 

## Notes

I didn't really test this yet.

## Test Cases

- Add rules for client and remove all rules for said client, then try to add rules for that client.

_If you did any manual testing (e.g., with Postman), put screenshots of the http request and before and after of the db_

## To Do

- [ ] Test properly

## Checklist

- [ ] No merge conflicts
- [ ] All checks passing

Closes #475

### DIFF
--- a/scylla-server/src/rule_structs.rs
+++ b/scylla-server/src/rule_structs.rs
@@ -368,7 +368,10 @@ impl RuleManager {
             }
             let clients = clients_op.unwrap();
 
-            let triggered = match self.rules.write().await
+            let triggered = match self
+                .rules
+                .write()
+                .await
                 .get_mut(&rule_id)
                 .map(|rule| rule.tick(&data.values))
             {
@@ -408,7 +411,10 @@ impl RuleManager {
     /// Adds a rule, creating or activating the client if needed
     pub async fn add_rule(&self, client: ClientId, rule: Rule) -> Result<(), RuleManagerError> {
         // Add to subscriptions bimap
-        self.subscriptions.write().await.insert(&client, &rule.id.clone());
+        self.subscriptions
+            .write()
+            .await
+            .insert(&client, &rule.id.clone());
 
         // Add to topic index
         self.topic_index
--- a/scylla-server/src/rule_structs.rs
+++ b/scylla-server/src/rule_structs.rs
@@ -40,7 +40,6 @@ pub struct BiMultiMap<L, R> {
     right_to_left: FxHashMap<R, FxHashSet<L>>,
 }
 
-#[allow(dead_code)]
 impl<L: Hash + Eq + Clone, R: Hash + Eq + Clone> BiMultiMap<L, R> {
     pub fn new() -> Self {
         Self {
@@ -355,59 +354,38 @@ impl RuleManager {
         let rule_ids = match self.topic_index.read().await.get(&data.name) {
             Some(rule_ids) => rule_ids.clone(), // Clone so we can drop resource
             None => {
-                trace!("Could not find rule in topic -> rule index: {}", data.name);
+                warn!("Could not find rule in topic -> rule index: {}", data.name);
                 return Err(RuleManagerError::NoMatchingRule);
             }
         };
 
         let mut notifications: Vec<(ClientId, RuleNotification)> = Vec::new();
         for rule_id in rule_ids {
-            let (triggered_result, clients_result) = {
-                // Future for if rule was triggered
-                let triggered_future = async {
-                    if let Some(rule) = self.rules.write().await.get_mut(&rule_id) {
-                        if let Some(triggered) = rule.tick(&data.values) {
-                            Ok(triggered)
-                        } else {
-                            Err(RuleManagerError::RuleFailure)
-                        }
-                    } else {
-                        trace!("Could not find rule in rules map: {}", rule_id);
-                        Err(RuleManagerError::NoMatchingRule)
-                    }
-                };
+            let clients_op = self.subscriptions.read().await.get_left(&rule_id).cloned();
 
-                // Future for getting subscribed clients
-                let clients_future =
-                    async { self.subscriptions.read().await.get_left(&rule_id).cloned() };
+            if clients_op.is_none() {
+                continue;
+            }
+            let clients = clients_op.unwrap();
 
-                tokio::pin!(triggered_future);
-                tokio::pin!(clients_future);
-
-                // Check which operation finished first
-                tokio::select! {
-                    triggered_result = &mut triggered_future => {
-                        match triggered_result? {
-                            true => (Ok(true), clients_future.await),
-                            false => (Ok(false), None),
-                        }
-                    },
-                    clients_result = &mut clients_future => {
-                        match clients_result {
-                            Some(_) => (triggered_future.await, clients_result),
-                            None => (Ok(false), None)
-                        }
-                    }
+            let triggered = match self.rules.write().await
+                .get_mut(&rule_id)
+                .map(|rule| rule.tick(&data.values))
+            {
+                Some(Some(val)) => val,
+                // Rule tick failed
+                Some(None) => return Err(RuleManagerError::RuleFailure),
+                None => {
+                    warn!("Could not find rule in rules map: {}", rule_id);
+                    return Err(RuleManagerError::NoMatchingRule);
                 }
             };
 
-            let triggered = triggered_result?;
-            if !triggered || clients_result.is_none() {
+            if !triggered {
                 continue;
             }
 
-            // Push notifications for all clients who are subscribed to this rule
-            for client in clients_result.unwrap() {
+            for client in clients {
                 notifications.push((
                     client.clone(),
                     RuleNotification {
@@ -429,29 +407,19 @@ impl RuleManager {
 
     /// Adds a rule, creating or activating the client if needed
     pub async fn add_rule(&self, client: ClientId, rule: Rule) -> Result<(), RuleManagerError> {
-        let rule_id = rule.id.clone();
-        let topic = rule.topic.clone();
+        // Add to subscriptions bimap
+        self.subscriptions.write().await.insert(&client, &rule.id.clone());
 
-        // Run all three writes concurrently
-        tokio::join!(
-            async {
-                // Add to subscriptions bimap
-                self.subscriptions.write().await.insert(&client, &rule_id);
-            },
-            async {
-                // Add to topic index
-                self.topic_index
-                    .write()
-                    .await
-                    .entry(topic)
-                    .or_insert(FxHashSet::default())
-                    .insert(rule_id.clone());
-            },
-            async {
-                // Add to rules lookup
-                self.rules.write().await.insert(rule_id.clone(), rule);
-            }
-        );
+        // Add to topic index
+        self.topic_index
+            .write()
+            .await
+            .entry(rule.topic.clone())
+            .or_insert(FxHashSet::default())
+            .insert(rule.id.clone());
+
+        // Add to rules lookup
+        self.rules.write().await.insert(rule.id.clone(), rule);
 
         Ok(())
     }
--- a/scylla-server/src/rule_structs.rs
+++ b/scylla-server/src/rule_structs.rs
@@ -49,6 +49,14 @@ impl<L: Hash + Eq + Clone, R: Hash + Eq + Clone> BiMultiMap<L, R> {
         }
     }
 
+    pub fn lefts(&self) -> Vec<L> {
+        self.left_to_right.keys().cloned().collect()
+    }
+
+    pub fn rights(&self) -> Vec<R> {
+        self.right_to_left.keys().cloned().collect()
+    }
+
     pub fn get_right(&self, left: &L) -> Option<&FxHashSet<R>> {
         self.left_to_right.get(left)
     }
@@ -497,5 +505,9 @@ impl RuleManager {
             .into_iter()
             .map(|rule| rule.clone())
             .collect()
+    }
+
+    pub async fn get_all_clients(&self) -> Vec<ClientId> {
+        self.subscriptions.read().await.lefts()
     }
 }
--- a/scylla-server/src/rule_structs.rs
+++ b/scylla-server/src/rule_structs.rs
@@ -10,7 +10,10 @@ use rustc_hash::FxHashSet;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use serde_with::DurationSeconds;
+use std::borrow::Borrow;
+use std::hash::Hash;
 use std::time::Duration;
+use tokio::sync::RwLock;
 use tracing::trace;
 use tracing::warn;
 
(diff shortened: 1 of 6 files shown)


suryatho:
I pushed my initial changes for the RuleManager, but I wasn't too sure how I should go about testing it (because I don't know how to create rules). It would be nice if you could tell me how you would like me to test it. It builds tho

bracyw:
> I pushed my initial changes for the RuleManager, but I wasn't too sure how I should go about testing it (because I don't know how to create rules). It would be nice if you could tell me how you would like me to test it. It builds tho

Definitely will, I'll plan to spend some time going over in Thursday meeting. For now if you want to spend some time outside of some proper guidance on this, look into using postman to test adding/deleting rules (via http) and subscribing to rules via the socket output. You should be able to use a fake client_id.

suryatho:
oh sounds cool, I will try to figure out how to make my own rules and send them using Postman. It doesn't seem too hard we even have a gwee for it.

One more question about the logic for the RuleManager:
I've written it up such that whenever you call `delete_rule(client, rule)` (for deleting a rule from a client), if the client no longer has any more rules then it is removed from `subscriptions` entirely, and if the rule is no longer serving anybody it also gets kicked out the data structure as a whole. Similarly when we `delete_client(client)` (for removing a client as a key from subscriptions), if any of the rules it was subscribed to no longer serve any clients, those rules also get kicked from the data structure. 

I forgot if you said something else which is why I am asking this question, but this was the way I figured it should be implemented.

So basically the question is: Whenever a rule no longer serves any clients, should it be deleted from all parts of the data structure? (Or should we keep it in the `topic_index` or `rules` maps) And if a client no longer is asking for any rules after a delete call should that client be removed from `subscriptions`? (It should right)

bracyw:
> oh sounds cool, I will try to figure out how to make my own rules and send them using Postman. It doesn't seem too hard we even have a gwee for it.
> 
> One more question about the logic for the RuleManager: I've written it up such that whenever you call `delete_rule(client, rule)` (for deleting a rule from a client), if the client no longer has any more rules then it is removed from `subscriptions` entirely, and if the rule is no longer serving anybody it also gets kicked out the data structure as a whole. Similarly when we `delete_client(client)` (for removing a client as a key from subscriptions), if any of the rules it was subscribed to no longer serve any clients, those rules also get kicked from the data structure.
> 
> I forgot if you said something else which is why I am asking this question, but this was the way I figured it should be implemented.
> 
> So basically the question is: Whenever a rule no longer serves any clients, should it be deleted from all parts of the data structure? (Or should we keep it in the `topic_index` or `rules` maps) And if a client no longer is asking for any rules after a delete call should that client be removed from `subscriptions`? (It should right)

It's looking good! 

(for commits can you try and model them after some of the closed PR's like single sentence with `#XXX` (ticket number in front)

For the first part I don't think we have a GUI, Argos 2 (a flutter a app Jack made) does have a GUI I believe.

For the deletion how bout we have a force delete on empty list argument  (boolean true if you want to auto delete, false if not) (so you can control whether or not you want to force delete only when you want to).

suryatho:
Oh I thought I had put #475 in each of the commit messages, I can do that then.

I meant like that postman has a gui for the making of these requests. (it does right ?)

Where would the force delete be? Is it on the `BiMultiMap` api such that it can leave keys with no values. Or do you mean on the api for RuleManager where we can leave rules that no longer serve any clients in the `topic_index` and `rules` maps.

bracyw:
> I meant like that postman has a gui for the making of these requests. (it does right ?)

oh yeah postman does, lmao (one day you will have to try out Argos 2, it's in flutter)


just in the BiMultiMap and for now the RuleManager won't delete, unless specifically asked (for a rule). So that way unsubscribing won't delete a rule auto. And for now we won't support for deleting a rule through unsubscribing client side.

suryatho:
> just in the BiMultiMap and for now the RuleManager won't delete, unless specifically asked (for a rule). So that way unsubscribing won't delete a rule auto. And for now we won't support for deleting a rule through unsubscribing client side.

So using this logic: there will never be mappings in the `subscriptions` bimap where a client is mapped to an empty set of rules and vice versa, but we will allow rules to exist in `topic_index` and `rules` without serving any clients (not in `subscriptions`). So basically from the client side we can always add rules but never delete them from the data structure entirely (we remove can from subscriptions). 

that makes sense!

I may have read what you meant wrong. But if I didn't, the solution is probably pretty easy, and that we never actually delete from the rule_cache and I can get rid of that code (Or should I keep it as part of the API for future?).

suryatho:
Oh when I was committing before the # in the message was being interpreted as a comment in the message.

bracyw:
> > just in the BiMultiMap and for now the RuleManager won't delete, unless specifically asked (for a rule). So that way unsubscribing won't delete a rule auto. And for now we won't support for deleting a rule through unsubscribing client side.
> 
> So using this logic: there will never be mappings in the `subscriptions` bimap where a client is mapped to an empty set of rules and vice versa, but we will allow rules to exist in `topic_index` and `rules` without serving any clients (not in `subscriptions`). So basically from the client side we can always add rules but never delete them from the data structure entirely (we remove can from subscriptions).
> 
> that makes sense!
> 
> I may have read what you meant wrong. But if I didn't, the solution is probably pretty easy, and that we never actually delete from the rule_cache and I can get rid of that code (Or should I keep it as part of the API for future?). I would just have to support adding rules that are already part of the cache more efficiently instead (upsert/don't insert if we already have a rule "cached").

Exactly what I was thinking for the unsub / sub functionality! As far as the rule_cach, deleting should still happen just not from a unsubscribing, only from an explicit rule deletion. (I guess what I was saying about sub/unsub for now doesn't matter because all the endpoints are oriented at a client adding or deleting rules (not unsubscribing which will come later).

suryatho:
I don't think we have an actual **_delete_** rule method on the `RuleManager` API in the sense that there is a method called `delete_rule`, but it was used previously (and still is) as a method to delete a rule from a client, or in other words, unsubscribe the client from a rule (unrelated but we do have `delete_client`). I don't think it would be fitting to delete the rule from the rules and topic cache here since this method seems like its only meant to remove a relationship of one client to one of it's rules. Do you suggest I write a real delete rule method that actually removes the rule data entirely from the RuleManager.

Maybe I am missing something here about the final architecture of the rule system, but wouldn't it be easiest to implement if the action of a client unsubscribing from a rule could lead to the deletion of that rule if no more clients are subscribed to that rule, and if the client disconnects or is deleted the rules it was subscribed to could be deleted if they aren't subscribed to by other clients.

bracyw:
> I don't think we have an actual **_delete_** rule method on the `RuleManager` API in the sense that there is a method called `delete_rule`, but it was used previously (and still is) as a method to delete a rule from a client, or in other words, unsubscribe the client from a rule (unrelated but we do have `delete_client`). I don't think it would be fitting to delete the rule from the rules and topic cache here since this method seems like its only meant to remove a relationship of one client to one of it's rules. Do you suggest I write a real delete rule method that actually removes the rule data entirely from the RuleManager.
> 
> 
> 
> Maybe I am missing something here about the final architecture of the rule system, but wouldn't it be easiest to implement if the action of a client unsubscribing from a rule could lead to the deletion of that rule if no more clients are subscribed to that rule, and if the client disconnects or is deleted the rules it was subscribed to could be deleted if they aren't subscribed to by other clients. 

The idea I was going for is that a client can unsubscribe from a rule and still go back to it later. If they really don't want it then they can delete. But transferring the logic is all that is required for this PR (so whatever is there logic wise we should keep (in terms of function))

suryatho:
I added a test file and wrote in some unit tests (it passes all tests), which was almost all AI but there were some important tests in there, notably: concurrent adding of rules, and testing of the handle_msg method in cases where there are mutiple clients as well. Let me know how much I should trim down the test file. I also had to make some changes in `rule_structs.rs` regarding the access levels of certain members.

scylla-server/tests/rule_structs_test.rs
```diff
@@ -0,0 +1,789 @@
...
+        run_id: 1,
+        name: "test/topic".to_string(),
+        unit: "test_unit".to_string(),
+        values: vec![15.0], // a = 15.0 > 10, should trigger
+        timestamp: Utc::now(),
+    };
+
+    // First trigger might not fire due to debounce logic
+    let _ = rule_manager.handle_msg(&client_data).await;
+
```
bracyw, line 110: This should test that there is no notification returned (and maybe just make debounce and wait time a little longer to ensure test isn't flaky (i.e. can fail sometimes depending on speed of computer))

bracyw: IK test under get's to this but probably should check here also for clarity

scylla-server/tests/rule_structs_test.rs
```diff
@@ -0,0 +1,789 @@
...
+        });
+        handles.push(handle);
+    }
+
+    // Wait for all tasks to complete
+    for handle in handles {
+        handle.await.unwrap()?;
+    }
+
+    assert_eq!(rule_manager.get_all_rules().await.len(), 10);
```
bracyw: should check that all were added, plus length.

scylla-server/tests/rule_structs_test.rs
```diff
@@ -0,0 +1,789 @@
...
+    let topics: Vec<_> = rules.iter().map(|rule| &rule.topic.0).collect();
+    assert!(topics.contains(&&"topic1".to_string()));
+    assert!(topics.contains(&&"topic2".to_string()));
+    assert!(topics.contains(&&"topic3".to_string()));
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_rule_manager_concurrent_operations() -> Result<(), RuleManagerError> {
```
bracyw: can you make a couple more of these tests

bracyw (commented):
Looks good, added a few comments on tests. No need to delete any of them, however maybe put bi-multi-map tests either at end of file or in a different file.

suryatho:
let me know what I should remove from the test files they're lowk very long.

suryatho:
I can't tell if my latest commit is like bad code or not but I was thinking you could do the stuff in handle_msg as an unordered set of tasks so I wanted to implement that. But that is also relying on the fact that we can read from rule managers from multiple tasks so the code looks weird? (or its fine idrk).

bracyw:
> I can't tell if my latest commit is like bad code or not but I was thinking you could do the stuff in handle_msg as an unordered set of tasks so I wanted to implement that. But that is also relying on the fact that we can read from rule managers from multiple tasks so the code looks weird? (or its fine idrk).

I would say the idea isn't bad (spawning multiple tasks to increase concurrent processing), but for now I think we should keep things relatively simple, because I don't think processing time will be a problem, rather possible bugs with Rwlocks which will be more problematic with multiple task spawning within the handle message function. After you switch back I'll be giving this a final review after Tuesday (I got a few things at work and stuff that I need to focus on until then.)

After this we can finalize our endpoints and frontend design, super exciting!

scylla-server/src/rule_structs.rs
```diff
@@ -10,17 +10,173 @@ use rustc_hash::FxHashSet;
...
+    RemovedOnly,
+    NothingToRemove,
+}
+
+pub struct BiMultiMap<L, R> {
+    left_to_right: FxHashMap<L, FxHashSet<R>>,
+    right_to_left: FxHashMap<R, FxHashSet<L>>,
+}
+
+#[allow(dead_code)]
```
bracyw: shouldn't silence warnings, either comment out / delete unused functions

scylla-server/src/rule_structs.rs
```diff
@@ -169,57 +340,86 @@ impl Default for RuleManager {
...
-        let topic = Topic(data.name.clone());
-
-        let Some(clients) = self.rules_lookup.get(&topic) else {
-            trace!("(normal) Could not find rule in rule cache: {}", data.name);
-            return Err(RuleManagerError::NoMatchingRule);
+        // Read from topic to rule index and drop lock immediately
+        let rule_ids = match self.topic_index.read().await.get(&data.name) {
+            Some(rule_ids) => rule_ids.clone(), // Clone so we can drop resource
+            None => {
+                trace!("Could not find rule in topic -> rule index: {}", data.name);
```
bracyw: can remove this trace

scylla-server/src/rule_structs.rs
```diff
@@ -169,57 +340,86 @@ impl Default for RuleManager {
...
-                                values: data.values.clone(),
-                                time: data.timestamp,
-                            },
-                        ));
+        let mut notifications: Vec<(ClientId, RuleNotification)> = Vec::new();
+        for rule_id in rule_ids {
+            let (triggered_result, clients_result) = {
+                // Future for if rule was triggered
+                let triggered_future = async {
+                    if let Some(rule) = self.rules.write().await.get_mut(&rule_id) {
```
bracyw: We will end up having to switch to a better data structure for rules, especially if there are performance issues–like you said before–but for now I think we can just delete the futures as they have to wait for the write lock anyway and instead just check before this if there are any clients attached to the rule to avoid the rules write lock unless there are subscribers. 

TLDR: remove futures and just wait for the locks, move clients check above rules check and abort check of rule if no one is subbed.

scylla-server/src/rule_structs.rs
```diff
@@ -169,57 +340,86 @@ impl Default for RuleManager {
...
+
+                // Future for getting subscribed clients
+                let clients_future =
+                    async { self.subscriptions.read().await.get_left(&rule_id).cloned() };
+
+                tokio::pin!(triggered_future);
+                tokio::pin!(clients_future);
+
+                // Check which operation finished first
+                tokio::select! {
```
bracyw: this wouldn't be needed if we delete those futures.

scylla-server/src/rule_structs.rs
```diff
@@ -228,105 +428,86 @@ impl RuleManager {
...
+            async {
+                // Add to subscriptions bimap
+                self.subscriptions.write().await.insert(&client, &rule_id);
             },
-
-            None => {
-                let mut map = FxHashMap::default();
-                map.insert(rule.topic.clone(), vec![rule]);
-                self.clients_map.insert(client, map);
+            async {
```
bracyw: this can also just be sequential for now. No need for async.

scylla-server/tests/rule_structs_test.rs
```diff
@@ -0,0 +1,544 @@
...
+    assert!(empty.is_ok_and(|op| op.is_none()));
+
+    // Wait for debounce
+    tokio::time::sleep(tokio::time::Duration::from_millis(150)).await;
+
+    let result = rule_manager.handle_msg(&client_data).await?;
+
+    if let Some(notifications) = result {
+        // Both rules should trigger since 15.0 > 10 and 15.0 > 5
+        assert!(notifications.len() >= 1);
```
bracyw: this is a little flaky because they should either both always trigger or not, so maybe an assert eq

scylla-server/tests/rule_structs_test.rs
```diff
@@ -0,0 +1,544 @@
...
+    for (value, result) in final_results {
+        assert!(
+            result.is_ok(),
+            "Failed to process message with value {}",
+            value
+        );
+
+        if let Ok(Some(notifications)) = result {
+            // Count how many rules should trigger for this value
+            let expected_triggers = num_rules - (value as usize / 10).min(num_rules);
```
bracyw: This I believe is wrong should be this:

      let expected_triggers = ((value as usize / 10).min(num_rules - 1)) + 1;

bracyw (changes requested):
couple small changes. lmk if you have questions

scylla-server/src/rule_structs.rs
suryatho: i think they are supposed to be warn! 's. Thats what its been like in other spots where we return errors

bracyw (approved):
LGTM, nice work! Feel free to merge.

PR #542, opened 2026-03-13
eFuses Page

bjackson312006:
New page for controlling and monitoring the eFuses on VCU 25A.

Here are the different bits of the UI:
Every eFuse has its own card that looks like this. They're titled with the name of the eFuse, as well as that eFuse's max current configuration.
image
These cards exist as a component, and you can configure a bunch of stuff about each card via the component parameters (e.g.., the eFuse name, max current, and all the specific topics the card needs to source its data from).

The "Current", "ADC Votlage", and "ADC" displays, as well as the "FAULTED" and "ENABLED" indicators, are pretty straightforward. They just constantly display those values as reported by the CAN messages being sent by VCU.

The "ON/OFF" button switches below this allow you to switch between the different possible states of the eFuse. When you click on any of these buttons, Calypso dispatches a command to VCU with the new state. Calypso is supposed to be the source of truth for those commanded states (i.e., VCU is purely a slave/follower in this setup). Importantly, the UI states of the buttons (i.e., which button is pressed down) is also set by the Calypso command messages, so it isn't just a local thing. This means that if two users have Argos open at the same time, and one of them clicks on one of the buttons to change an eFuse state, that change will also come through on the second user's screen. Because the UI states of the buttons is set by Calypso command messages, the buttons go kinda crazy when you're in sim mode (they randomly switch states on their own as mock data comes in). Can't really get around that though.

Below the eFuse state buttons, there's also a smaller display that displays the selected state for that eFuse, but as according to VCU. Basically, when VCU receives a state command from Calypso, it updates that state within its own internal eFuse state tracking system. Then, as part of the eFuse messages, VCU reports what it is currently storing for that value. This value is what's shown in the smaller display. This is mainly useful for debugging (i.e., we can see how quickly VCU receives state commands from Calypso and actually updates the eFuse state in the firmware). If everything is working correctly, a user would click on one of the buttons to send a Calypso command, and then after a short delay, you'd see the display update to match this new state as VCU receives the command and updates the state internally.

Also, certain cards have an additional control state called "AUTO". For example, the Radfan eFuse card looks like this:
image
The "AUTO" state just tells VCU to automatically turn the eFuse on or off based on certain temp readings in the firmware. For the eFuses that support this mode, there's an additional display off to the right of the buttons that shows the relevant temperature value being used by VCU to determine the on/off state. These displays show the temperature values as understood by VCU, rather than the DTI or BMS. So, even though the temperature values come from the DTI/BMS (instead of VCU), these displays show VCU's echoes of these values after parsing them from other boards in the firmware. This is helpful because we can directly know the exact values VCU is actually using in the firmware for automatic state control.

Other stuff:
- I added two new fonts (Highway Gothic, and the seven segment display font). Highway Gothic is public domain cus it's a government font, but the seven segment display font is just something i took from a [reddit comment](https://www.reddit.com/r/Woovebox/comments/1e26ofo/7_segment_lcd_font/). The guy said that it has an open font license tho
- I added two new compse ymls for my local testing, they both use my local calypso builds so I should probably remove before merging. But for now i'm gonna keep them in case I need to do more testing


angular-client/src/pages/efuses-page/components/efuse-card/efuse-card.component.html
```diff
@@ -0,0 +1,112 @@
...
+
+        <!-- ADC Voltage + ADC Raw stacked vertically -->
+        <vstack [spacing]="'10px'" [align]="'flex-start'">
+          <!-- ADC Voltage -->
+          <vstack [align]="'flex-start'">
+            <seven-segment-display
+              [value]="voltage.toFixed(2)"
+              [digits]="3"
+              [decimals]="2"
+              [fontSize]="40"
```
bracyw: can you make the inputs to seven-segment that are shared variables (e.g. small fontSize for ADC)

angular-client/src/data-type.enum.ts
```diff
@@ -26,6 +26,12 @@ export enum DataTypeEnum {
   // DTI
   MOTOR_TEMP = 'DTI/Temps/Motor_Temperature',
 
+  // VCU Echo (AUTO-mode source-of-truth)
+  VCU_ECHO_MOTOR_TEMP = 'VCU/Echo/Motor_Temp',
```
bracyw: Can you move these to topic.utils.ts and maybe make utils functions for applicable functions

bracyw (changes requested):
lgtm couple changes

bracyw (approved):

bracyw:
🔥

PR #602, opened 2026-04-20
#600 Handle broadcast RecvError in handling_loop and socket_handler

### LINKED ISSUE #600: Fix silent hang in db_handler handling_loop when broadcast sender closes
### Observed Behavior

handling_loop in db_handler stopped producing batches on 2026-04-20 at 02:49:30.864 UTC and emitted zero log output for ~2 hours (no info, warn, error, or panic) until scylla-server was restarted. Socket fanout and MQTT ingress kept running at ~16 msg/sec the entire time. pg_stat_activity showed only idle scylla connections, and bb8 pool.get() never timed out — ruling out pool exhaustion and DB locks.

Root cause at scylla-server/src/db_handler.rs:184 — the select! arm only matches Ok:

```rust
Ok(msg) = self.receiver.recv() => { self.handle_msg(msg, &data_channel).await; }
```

When the broadcast sender is dropped, self.receiver.recv() returns Err(RecvError::Closed) on every poll. The pattern fails, the branch is disabled for that select! iteration, and no log is emitted. The loop re-enters and hits the same error on the next iteration. data_queue stays empty so batch_interval.tick() skips the send. The task spins silently forever.

The same unhandled pattern exists at socket_handler.rs:38 and socket_handler.rs:156.

Pattern introduced 2024-11-28 (commit 8ff34865), latent for ~17 months. Suspected trigger for the 2026-04-20 incident is PR #539 (rumqttc git-main → 0.25.1 repin, tokio 1.48 → 1.49, edition 2024) — likely a reconnect or error path in the new rumqttc version drops the mqtt_send_db sender clone while process_mqtt keeps running on mqtt_send_socket.

### Expected Behavior

- Err(RecvError::Closed) flushes any remaining data_queue and exits the loop cleanly.
- Err(RecvError::Lagged(n)) emits a warn with the skip count and continues the loop (broadcast auto-recovers on next recv).
- Every error variant is observable in logs; silent cases are not acceptable for a long-running task.

### Steps to Reproduce

1. See the e2e test at scylla-server/tests/db_handler_test.rs.
2. It spawns handling_loop with a real pool, drops the sole broadcast sender, and asserts the task does not exit within 3 batch_interval windows and emits no batches.
3. Bring up the backend (./argos.sh scylla-dev up) and run cargo test --test db_handler_test.
4. With current code the test passes, documenting the bug. After the fix the test will fail and must be inverted to assert task.is_finished().

### Proposed Fix

Replace the select! arm at db_handler.rs:184 with explicit error handling:

```rust
recv_result = self.receiver.recv() => {
    match recv_result {
        Ok(msg) => self.handle_msg(msg, &data_channel).await,
        Err(broadcast::error::RecvError::Lagged(n)) => {
            warn!("db_handler broadcast lagged, skipped {} messages", n);
        }
        Err(broadcast::error::RecvError::Closed) => {
            warn!("db_handler broadcast closed; flushing and exiting");
            if !self.data_queue.is_empty() {
                let _ = data_channel.send(self.data_queue).await;
            }
            break;
        }
    }
}
```

Apply the same treatment to socket_handler.rs:38 and socket_handler.rs:156. Invert the assertion in tests/db_handler_test.rs to expect the task to finish after the sender is dropped. Consider a follow-up to confirm the rumqttc 0.25.1 drop path that triggered the production incident.

bracyw:
## Changes

Replaces the Ok(msg) = self.receiver.recv() select! arm in db_handler.rs:184 with a full match on the Result, so broadcast RecvError variants are handled explicitly instead of silently disabling the branch. Applies the same treatment to both socket_handler variants at src/socket_handler.rs:38 and :156. Adds an e2e integration test that spawns handling_loop against a real Postgres pool, sends messages, drops the broadcast sender, and verifies the pending batch is flushed and the task exits cleanly.

- db_handler: Err(Lagged) warns with skip count and continues; Err(Closed) warns, flushes data_queue through data_channel, and breaks the loop
- socket_handler (both variants): Err(Lagged) warns and continues; Err(Closed) warns and breaks
- New integration test covers happy-path insertion, interval-isolated flush, and exit-on-close

## Notes

The bug was latent since 2024-11-28 (commit 8ff34865) when the receiver type switched from mpsc to broadcast. Before this PR, any Err return from self.receiver.recv() caused the Ok(msg) pattern to silently fail — the select! branch was disabled for that iteration, no log was emitted, and the loop spun forever on the interval/cancel branches with zero observable output. This matches the 2026-04-20 production incident where handling_loop went silent for ~2 hours.

This PR fixes the symptom. It does not prove what originally dropped the mqtt_send_db sender in production — PR #539's rumqttc 0.25.1 repin is the strongest suspect, worth a separate follow-up. With this fix in, any recurrence will at minimum log the exit instead of going silent.

Test scope: handling_loop linked as a library against a real Postgres pool, but bypassing process_mqtt, batching_loop, and rumqttc. A fuller e2e test would spin up the whole task graph; deferred.

## Test Cases

- handling_loop drains buffered broadcast messages before exit and flushes the pending data_queue once via data_channel
- Task exits within 500ms of broadcast sender drop, proving Err(Closed) is handled
- Interval-tick flush path cannot race the close-handling flush (BATCH_UPSERT_TIME set to u16::MAX in the test, isolating exit to the Err(Closed) path)
- All pre-existing scylla-server integration tests still pass (45 tests)

## Checklist

- [x] No merge conflicts
- [ ] All checks passing

Closes #600

### DIFF
--- a/scylla-server/src/db_handler.rs
+++ b/scylla-server/src/db_handler.rs
@@ -181,8 +181,20 @@ impl DbHandler {
                     data_channel.send(self.data_queue).await.expect("Could not comm data to db thread, shutdown");
                     break;
                 },
-                Ok(msg) = self.receiver.recv() => {
-                    self.handle_msg(msg, &data_channel).await;
+                recv_result = self.receiver.recv() => {
+                    match recv_result {
+                        Ok(msg) => self.handle_msg(msg, &data_channel).await,
+                        Err(broadcast::error::RecvError::Lagged(n)) => {
+                            warn!("db_handler broadcast lagged, skipped {} messages", n);
+                        }
+                        Err(broadcast::error::RecvError::Closed) => {
+                            warn!("db_handler broadcast closed; flushing and exiting");
+                            if !self.data_queue.is_empty() {
+                                let _ = data_channel.send(self.data_queue).await;
+                            }
+                            break;
+                        }
+                    }
                 }
                 _ = batch_interval.tick() => {
                     if !self.data_queue.is_empty() {
--- a/scylla-server/src/socket_handler.rs
+++ b/scylla-server/src/socket_handler.rs
@@ -35,8 +35,17 @@ pub async fn socket_handler(
                 debug!("Shutting down socket handler!");
                 break;
             },
-            Ok(data) = data_channel.recv() => {
-                send_socket_msg(&data, &mut upload_counter, &io, DATA_SOCKET_KEY).await;
+            recv_result = data_channel.recv() => {
+                match recv_result {
+                    Ok(data) => send_socket_msg(&data, &mut upload_counter, &io, DATA_SOCKET_KEY).await,
+                    Err(broadcast::error::RecvError::Lagged(n)) => {
+                        warn!("socket_handler broadcast lagged, skipped {} messages", n);
+                    }
+                    Err(broadcast::error::RecvError::Closed) => {
+                        warn!("socket_handler broadcast closed; exiting");
+                        break;
+                    }
+                }
             }
         }
     }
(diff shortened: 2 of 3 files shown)


bracyw:
no longer seems to be an issue

bracyw:
keeping branch around in case the issue comes back

PR #609, opened 2026-04-22
Enable Certain eFuses

bjackson312006:
This PR makes some changes to the eFuses page, since all eFuses are now able to be enabled/disabled again by VCU firmware. I decided to split the eFuses into two sections: A "Normal" section, and a "Caution" section.
- The eFuses in the "Normal" section can easily be toggled from Argos with nothing bad happening. You shouldn't really need to mess with their states from Argos during normal operation, but you can if you want/need to.
- The LV, Shutdown, Battbox, and MC eFuses are in the "Caution" section. Since these eFuses control certain more-critical functions in the car, you have to "unlock" them to override their state. There's a notice above the section stating that these eFuses shouldn't really have their states messed with during normal operation, aside from an emergency. In particular, the Shutdown and LV eFuses will literally turn off TPU, so if you disable them via Argos, you'd have to physically go back up to the car to reset it.

### DIFF
--- a/angular-client/src/pages/efuses-page/components/efuse-card/efuse-card.component.ts
+++ b/angular-client/src/pages/efuses-page/components/efuse-card/efuse-card.component.ts
@@ -13,7 +13,11 @@ import EfuseSwitchComponent, { EfuseSwitchState } from '../efuse-switch/efuse-sw
 import { LockButtonComponent } from '../lock-button/lock-button.component';
 import { inject } from '@angular/core';
 
-export type EfuseLockMode = 'Unlocked' | 'Unlockable' | 'Locked';
+export enum EfuseLockMode {
+  Unlocked,
+  UnlockToEdit,
+  ReadOnly
+}
 
 /**
  * Component to display individual eFuse status.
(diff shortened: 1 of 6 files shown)


angular-client/src/pages/efuses-page/efuses-page.component.css
bracyw: can you delete

angular-client/src/pages/efuses-page/components/efuse-card/efuse-card.component.ts
```diff
@@ -13,6 +13,8 @@ import EfuseSwitchComponent, { EfuseSwitchState } from '../efuse-switch/efuse-sw
 import { LockButtonComponent } from '../lock-button/lock-button.component';
 import { inject } from '@angular/core';
 
+export type EfuseLockMode = 'Unlocked' | 'Unlockable' | 'Locked';
```
bracyw: can you make this an enum and use it in the if statement checks, and inputs

angular-client/src/pages/efuses-page/components/efuse-card/efuse-card.component.ts
```diff
@@ -91,6 +93,13 @@ export default class EfuseCardComponent implements OnInit, OnDestroy {
   // ── Locking state ──
   isLocked = signal<boolean>(true);
 
+  effectivelyLocked = computed(() => {
```
bracyw: can you add this:
```
  private showsLockButton    = computed(() => this.lockMode() === EfuseLockMode.UnlockToEdit);
  private isForceLocked      = computed(() => this.lockMode() === EfuseLockMode.ReadOnly);                                                                
  private relockAfterCommand = computed(() => this.lockMode() === EfuseLockMode.UnlockToEdit);                                                            
                                                                                                                                                          
  effectivelyLocked = computed(() =>                                                                                                                      
    this.isForceLocked() ? true : this.isLocked()                                                                                                         
  );     
```

or something like it.

bracyw (changes requested):
couple changes

bracyw (approved):

Issue #697, opened 2026-07-13
Add S Volts cell-voltage heatmap view and LV battery tile


bracyw:
Tracking issue for the S voltages / LV battery spec. The spec is being drafted via to-spec and reviewed as a PR; on merge this issue is populated with the finalized spec and labeled ready-for-agent.

## Scope (draft)

- Add an S Volts per-cell heatmap view on the BMS debug page, mirroring the existing cell-voltage view (the S-ADC reading), and rename the current Voltage view to C Volts (the C-ADC reading).
- Add an LV battery voltage tile on the efuses page, reading VCU/LV/voltage, with a low-voltage warning driven by the LV_LOW_VOLTAGE_FAULT flag.

## Deferred

- Open Wire per-cell heatmap view — blocked on firmware pinning the OW payload (0/1 flag vs per-cell value); handled separately.

bracyw:
Hollow tracking stub. Superseded by #704 (S Volts view + rename) and #705 (LV Battery tile), filed directly on the tracker under the specs-only review gate (ADR 0006). Glossary lands via #701.

PR #698, opened 2026-07-13
Spec: S Volts heatmap view + LV battery tile

bracyw:
Spec (to-spec) for #697, staged for review per docs/agents/spec-review.md. Adds an S Volts (S-ADC) per-cell heatmap view mirroring the existing cell-voltage view, renames the current Voltage view to C Volts (C-ADC), and adds an LV battery voltage tile to the efuses page with a fault-driven low-voltage warning. Also adds the C-ADC / S-ADC / CvS glossary vocabulary to CONTEXT.md.

Frontend-only — telemetry ingest is dynamic, so no scylla-server, Charybdis, or Siren change.

Two firmware contracts to confirm before merge:
- The S_Volts per-cell topic string is not yet in the Odyssey definitions; confirm the exact string with the firmware owner or the tiles render grey.
- The LV low-voltage fault topic differs between the definitions (with Non-Critical) and the local simulator (without); confirm what the car publishes.

Open Wire per-cell view is deferred / out of scope pending the firmware payload shape.

On merge: the spec publishes to #697 with ready-for-agent, then to-tickets slices it (S Volts / LV battery; Open Wire separately). The staged docs/spec files are removed in a small follow-up PR once the tickets exist.

Refs #697

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -26,6 +26,20 @@ _Avoid_: Path, channel, stream.
 An on-vehicle CAN device that reports faults — one of `Bms`, `Dti`, `Mpu`, `Charger`. Used for attributing faults to their source.
 _Avoid_: Device, ECU, board, unit.
 
+### BMS cell voltages
+
+**C-ADC voltage** (**C Volts**):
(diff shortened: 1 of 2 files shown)


bracyw:
Superseded by two implementation tickets under #697; decisions + CONTEXT.md glossary carry into the ticket-set PR. Branch kept.

